PR #11, opened 2024-12-17
Merge wombatlite + cable 3 changes into dev/preindustrial+concentrations

blimlim:
This PR brings in the changes needed to run ESM1.6 using wombatlite and cable3 into the dev/preindustrial+concentrations branch.

Changes include: updated namelists, new input files, and new restart files.


atmosphere/cable.nml
pearseb, line 1: No idea if this is correct or not.

config.yaml
```diff
@@ -24,9 +24,9 @@ jobfs: 1500MB
 # Modules for loading model executables
 modules:
   use:
-      - /g/data/vk83/modules
+      - /g/data/vk83/prerelease/modules
   load:
-      - access-esm1p5/2024.05.1
+      - access-esm1p6/pr26-4
```
pearseb, line 29: We aren't going to pr27?

blimlim:
@pearseb, I just noticed that the `monthly 2d BGC surface fields` section of the diag table yaml files don't have the `module_name: generic_wombatlite` line that the other wombat sections have. Should this be added in?

https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/83ad4f1b8134c276e057eb0ac9e47bfe5a22b057/ocean/diagnostic_profiles/source_yaml_files/diag_table_standard_source.yaml#L357-L362

config.yaml
```diff
@@ -118,7 +117,7 @@ collate:
     walltime: 1:00:00
     mpi: false
 
-restart: /g/data/vk83/configurations/inputs/access-esm1p5/modern/pre-industrial/restart/
+restart: /g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2024.12.10
```
pearseb, line 120: does this restart have the latest restart file I sent you? The one from yesterday with ESM hack-a-thon?

make_wombatlite_restart.py
pearseb, line 1: This file is out of date and will not work with WOMBATlite. It was meant for the legacy WOMBAT.

config.yaml
blimlim: Good question, that will be the next step. This PR only brings in the wombatlite and cable 3 changes, matching the executables in pr26. 

The executables in pr27 need the iceberg files and namelist changes, and so I'll make all those changes after this.

config.yaml
blimlim: Yep! It has the `ocean_wombatlite.res.nc-20241211` file, (which is then renamed as `ocean_wombatlite.res.nc-01001231-v20241211`)

ocean/diagnostic_profiles/source_yaml_files/diag_table_standard_source.yaml
```diff
@@ -335,22 +336,31 @@ diag_table:
...
             output_freq: 1  # integer: output sampling frequency in output_freq_units (0: every timestep; -1: only at end of run)
             output_freq_units: months  # time units for output: years, months, days, hours, minutes, or seconds
         fields:
-            stf03:
-            stf07:
-            stf09:
-            stf10:
+            o2_stf:
+            dic_stf:
+            adic_stf:
```
pearseb, line 347: This isn't right, but fine for now. We will need to update this diagnostics file properly at a later date.

atmosphere/cable.nml
blimlim: Hopefully!

These Cable changes were also reviewed by @rml599gh in https://github.com/ACCESS-NRI/access-esm1.6-configs/pull/10

pearseb (approved):
Hi Spencer,
There are a few things that need correcting, but this can be done at a later date. It seems fine currently if we just want to get things running. I can also alter some of these files in my own local and push them back to the remote as I fix things.

ocean/diagnostic_profiles/source_yaml_files/diag_table_standard_source.yaml
blimlim: Thanks Pearse, I'll make an issue for fixing these and removing the `make_wombatlite_restart.py` file

PR #33, opened 2025-02-13
Increase collation walltime to 1.5h

### LINKED ISSUE #32: Increase collation walltime
@JhanSrbinovsky reports that during long simulations, the UM output is occasionally not properly converted to netCDF. In previous simulations, this appeared to arise from the collation job taking much longer than usual and running out of walltime. As the netCDF conversion is called from the collation, it doesn't run in these situations.

A local workaround of increasing the collation to 1.5h has been used, but it would be good to add this change to the configurations on github.

blimlim:
Closes #32 
This PR increases the collation walltime to 1.5h to try and prevent occasional jobs where the walltime is exceeded.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -116,7 +116,7 @@ collate:
     exe: mppnccombine.spack
     restart: true
     mem: 4GB
-    walltime: 1:00:00
+    walltime: 1:30:00
     mpi: false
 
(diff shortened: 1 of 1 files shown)


JhanSrbinovsky (approved):
LGTM

PR #42, opened 2025-02-25
Swap PI configuration to Sapphire Rapids queue

blimlim:
@manodeep found [here](https://github.com/ACCESS-NRI/ACCESS-ESM1.6/issues/45) that using the Sapphire Rapids queue and changing the processor layouts for the atmosphere and ocean reduced walltime by ~22% and reduced SU usage by ~10% in the pre-industrial configuration.

This PR adds @manodeep's changes to the configuration, namely:
- Swapping the queue to Sapphire Rapids, and adding the queue information to the `config.yaml`
- Changing the ocean decomposition 18x10 (180) -> 14x14 (196)
- Changing the atmosphere decomposition 16x12 (192) -> 16x13 (208)

As discussed in the [issue here](https://github.com/ACCESS-NRI/ACCESS-ESM1.6/issues/45), this change breaks bitwise reproducibility with the previous configuration. _Note_ this appears to be due to changing the decomposition, and not due to changing the hardware. Running with the old decomposition on the Sapphire Rapids queue reproduces the old results.


MartinDix:
The UM should be reproducible across different decompositions, but MOM in CM2 (and I think ESM1.5/6) was not (even when the build options were supposed to be reproducible).

aidanheerdegen (approved):
LGTM, but we should get @manodeep's approval I think.

manodeep:
@blimlim Should we re-do this with the updated UM7 (anything after [7288a51](https://github.com/ACCESS-NRI/UM7/commit/7288a51bf7aae3d2836aca674a1ae7ed15796e01) ) and verify? Might be worthwhile checking that the same exe produces identical results for UM, regardless of the decomposition.

blimlim:
That sounds like a good plan! I can put together a temporary deployment. We might have to wait while the gadi queue is having issues

manodeep:
@blimlim That sounds great! I want to run both the configs again and confirm the determinism before we recommend adopting the new configs.

blimlim:
@manodeep, I'm deploying some temporary executables [here](https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/60) which we can test with.

I've based it off the latest deployment from [PR42](https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/43) and updated the UM to [7288a51](https://github.com/ACCESS-NRI/UM7/commit/7288a51bf7aae3d2836aca674a1ae7ed15796e01). This does include some ocean changes to compared to our last run.

Since we know the PI-config won't reproduce across ocean decompositions, do you think we should just rerun the AMIP tests?

manodeep:
@blimlim That's great - thanks! I will re-run a set of `amip` configs based with the exes from that PR. I did notice that you moved some of the components and the spack-packages to older versions - is that intentional? The reason I ask is that (some/all? of) those later tags added support for oneAPI and being able to compare the same hashes between the classic and oneAPI compilers (down the track) would be a great bonus. 

Separately: Would the atmosphere component be deterministic regardless of the number of PEs even in the PI config?

blimlim:
>  I did notice that you moved some of the components and the spack-packages to older versions - is that intentional?

Good question. I'd based it off [PR43](https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/43) as we'd used it for the earlier tests. Would it be better to use something based on the updates in `main`? I can quickly put something together if so

> Separately: Would the atmosphere component be deterministic regardless of the number of PEs even in the PI config?

Yeah! I think if we varied the atm decomposition but not the ocean in the PI config, we'd expect results to stay the same

manodeep:
> > I did notice that you moved some of the components and the spack-packages to older versions - is that intentional?
> 
> Good question. I'd based it off [PR43](https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/43) as we'd used it for the earlier tests. Would it be better to use something based on the updates in `main`? I can quickly put something together if so
>

@blimlim Yes please - future-me will be very thankful 😄

> > Separately: Would the atmosphere component be deterministic regardless of the number of PEs even in the PI config?
> 
> Yeah! I think if we varied the atm decomposition but not the ocean in the PI config, we'd expect results to stay the same

That's great - I will run both the PI and AMIP configs and hopefully everything will be deterministic where expected. And then we can roll out the changes for swapping to the SPR queue.

blimlim:
Great! New exes will be built deployed [here](https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/62)

manodeep:
@blimlim How do we test for deterministic results in the PI config? The atmosphere is not the same between the different PE layouts I tried in the `sapphirerapids` queue (one I checked manually started to diverge at the 7th iteration)

manodeep:
Yes - the ocean layouts were different in the three cases that I ran. Should I run one with identical ocean config to check?

blimlim:
Ok that's good to know – unfortunately we expect the answers to change whenever the ocean decomposition changes, but it does mean those differences would be expected.

blimlim:
!test repro

blimlim:
!test repro commit

manodeep (approved):
Once the new checksums are committed, the changes can be merged in.

manodeep (approved):
New checksums are in

Issue #44, opened 2025-02-26
Default configuations to be added to cable.nml


JhanSrbinovsky:
cable_user%L_REV_CORR  = .TRUE. 
snmin = 0.11
cable_user%FWSOIL_SWITCH = 'Haverd2013'
cable_user%SOIL_THERMAL_FIX = .TRUE.

@blimlim  can you please add these switches to the default cable.nml

Actually this one is probably there as:
cable_user%FWSOIL_SWITCH = 'standard'
just needs to changed to 'Haverd2013'.

I have made these changes in the atmosphere/cable.nml of the payu control directory in a run that is currently going. From our previous discussions this should get picked up in the next year right?

FYI @har917 @rml599gh

blimlim:
Can do!

I've just created PRs for this 
#46 
and 
#45

PR #45, opened 2025-02-26
Update switch values in cable.nml

blimlim:
Closes PI half of #44 

Adds user switches with values listed in #44.


JhanSrbinovsky (approved):
LGTM

PR #46, opened 2025-02-26
Update switch values in cable.nml

blimlim:
Closes AMIP half of #44 

Adds cable user switches with values given in #44


JhanSrbinovsky (commented):
LGTM - Ta

JhanSrbinovsky (approved):
LGTM

PR #52, opened 2025-03-12
Use LUH3 restart

### LINKED ISSUE #39: Implement LUH3 based restart file as the default restart
@Whyborn has manufactured an LUH3 based restart file (PI-02-WithLUH3VegetationMap.astart)

This has been tested via a user-script work-around in dev-LUH3-restart-expt-5eed0c42/

Thorough analysis is yet to be performed. Perfunctory analysis/conclusion is that is fine.

It has been verified that the PFT distribution is in fact identical in PI-02-WithLUH3VegetationMap.astart restart000/atmosphere/restart_dump.astart is indeed AND in later years (for e.g. restart010/atmosphere/restart_dump.astart)

blimlim:
Closes #39 

Adds the new LUH3 based restart as the default for the pre-industrial configuration.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -119,7 +119,7 @@ collate:
     walltime: 1:30:00
     mpi: false
 
-restart: /g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2024.12.10
+restart: /g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2025.02.17
 
 # Timing controls
 calendar:
--- a/manifests/restart.yaml
+++ b/manifests/restart.yaml
@@ -1,192 +1,197 @@
 format: yamanifest
 version: 1.0
 ---
-work/atmosphere/PI-02.astart-01010101:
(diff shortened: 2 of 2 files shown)


JhanSrbinovsky:
@blimlim I haven't signed off yet because developments in the interim have suggested that the temperature drift in the ocean (that has emerged on the land developments side - not Pearse's drift) MAY be (at least partially) a result of the PFT map. 

There were concurrently the config changes in cable.nml l_rev_corr = .TRUE. that were made. Now this is correct. And the drift appeared  at the same time this did. We've been thinking that it was due to the code changes that have occurred. Current theory is that this drift (is perhaps exacerbated by the LUH3 PFT map). We are re-running a test with l_rev_corr switched on from the beginning. Previously l_rev_corr was not switched on until year=23. Ian is arguing that the damage was already done so we cant attribute blame.

All of that said, there is a chance that a third alternative PFT map will be adopted. We are having a meeting this afternoon, and hopefully will have some results by tomorrow.

JhanSrbinovsky (approved):
LGTM - Just a question - why do all the other files need to change? I'm guessing they are just in updated location?

blimlim:
Thanks @JhanSrbinovsky. The manifest files keep track of the restart files, inputs, and exes used by the configurations, storing their filepaths and hashes. Payu will update the manifest files automatically when running, but for the online configurations, we want the manifests to remain up to date.

Issue #40, opened 2025-02-25
Consolidate code and configuration with latest changes


ccarouge:
@har917 @JhanSrbinovsky I am suggesting we move the discussion that has erroneously started in [the AM3 PR](https://github.com/CABLE-LSM/ACCESS3-JULES/pull/269) into this issue.

Instead of me badly parroting you, if you could write the questions and workflow you have in mind here that would be great. It would make sure this discussion is visible beyond the 3-4 of us.

har917:
Between December 2024 and February 2025 there have been a number of developments within the CABLE3 codebase - these effect the science and the configuration (so choice of switches, source of ancillaries and other inputs) of the code.  To date development has been done in sections - not always sequence, some parallel devs have occurred - I understand that a run using all devs to date has recently been started.

We would like to rebase our existing long run of ESM1.6 onto this new codebase (alongside changes in the ocean model) - so the primary question is how should this be done?  What is the sequence of steps?  What level of testing/assessment is needed?  What constitutes success?  In particular, which configuration should form the baseline of any new runs and how does this relate to the existing long run?

Key sub-questions around baseline configuration (either for interim testing or for rebasing the long run):
- which source of soil maps to use?  The MOSES maps, current AM3-era maps, or the old constant soil properties maps in ESM1.5.
- which land cover map? Eventually we should switch to LUH3 maps but that maybe a secondary step. 
- which vegetation height map? 
- which rooting distribution parameters? which set of other vegetation parameters?
- which science configuration should be use?  As the current long ESM1.6+CABLE2.5 run or with other preferred tweaks. At the minimum the switches to consider are 'fwsoil', 'thermal_fix', 'medlyn'/'leuing', and 'access13roots'/'rootbeta'.  @rml599gh has highlighted the possibility os using/extending the hydraulic redistribution capability but it is likely that consideration of this is pushed to a future round.

Other points to consider:
- does this rebase occur before the wood thinning code + revised new initialisation is finalised?
- does this rebase occur before we have tested the land use change capability?
- how do ensure that any partner changes within payu are kept in sync?
- does this rebase occur before the development of the ACCESS-> iLAMB capability?

rml599gh:
Soil maps - I thought we already used MOSES type maps for most soil properties. AM3-era maps are probably worth trying but maybe as a parallel run.
Land cover - we should switch to the LUH3 based one unless anyone finds an issue with the test simulations in the next few days.
[...]

PR #54, opened 2025-03-14
Switch on LUC for AMIP run

### LINKED ISSUE #53: Switch on land use change for AMIP run
When copying the new cable namelist from the pre-industrial to the AMIP configuration, I erroneously left the land use change flag `l_luc` as `.false.`.

For the AMIP run, this should be `.true.`. This has caused some of the crashes when using the new vegetation input.

blimlim:
Closes #53 

Switches on `l_luc` for the AMIP configuration, as required for the land use change to run.

Compared to the AMIP namelist for ESM1.5, differences are:

```diff
diff esm1p5cable cable.nml 
3,5d2
<  filename%veg     = 'INPUT/def_veg_params.txt'
<  filename%soil    = 'INPUT/def_soil_params.txt'
< !
8a6
>   cable_user%GS_SWITCH = 'medlyn' ! alternative is 'leuning' - used for ESM1.5
15c13
<   cable_user%FWSOIL_SWITCH = 'standard' ! Controls root water uptake function
---
>   cable_user%FWSOIL_SWITCH = 'Haverd2013' ! Controls root water uptake function
34a34,39
>   cable_user%access13roots = .TRUE. ! TRUE uses prescribed froot values
>                                     ! FALSE uses 'rootbeta' parameterisation (as ACCESS-CM2)
>   cable_user%L_REV_CORR = .TRUE.
>   snmin = 0.11
>   cable_user%SOIL_THERMAL_FIX = .TRUE.
> 
58c63,64
<   casafile%cnpbiome='INPUT/pftlookup_csiro_v16_17tiles_wtlnds.csv'  ! biome specific BGC parameters
---
> ! updated pftlookup file for CABLE3 format
>   casafile%cnpbiome='INPUT/pftlookup_cable3.csv'  ! biome specific BGC parameters
```

Let me know if there's anything here not right for an amip simulation

### DIFF
--- a/atmosphere/cable.nml
+++ b/atmosphere/cable.nml
@@ -48,7 +48,7 @@
   satuParam = 0.8     ! only used if hydraulic redistribution .true.
 !
 ! CASA-CNP flags
-  l_luc = .FALSE.
+  l_luc = .TRUE.
   l_thinforest = .FALSE. 
   l_casacnp = .TRUE. ! redundant? actually controlled by icycle
   icycle = 3 ! Controls whether CASA-CNP is run and for which species


JhanSrbinovsky (commented):
My only query is a pretty fundamental one - did everyone land on wanting to turn this on yet? I thought @rml599gh, @Whyborn were playing with the 1850 map?

blimlim:
I think so – my understanding [from this comment](https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/40#issuecomment-2696363893) is that we want to use the AMIP configuration as a test bed for the restart file and vegetation changes, to find out if they work. 

The configuration originally had LUC switched on, using the CMIP6 vegetation, and I inadvertently switched it off when updating the namelist.

JhanSrbinovsky:
OK - So I guess you're right in the sense that we will need a configuration to test this in BUT was there an appetite for a general AMIP configuration? Do we have any AMIP configuration working? The assumption is that  there is a performance gain in eliminating the Ocean model. Given that it spends so much top in IO this might not be so dramatic a saving. At any rate the LUC itself is not free of performance penalty. i.e. "They" might've had in mind being able to do quick tests on other things as well for which LUC is an un-necessary burden. What do you think?

blimlim:
Good questions @JhanSrbinovsky. I'll try to clarify!
> was there an appetite for a general AMIP configuration? 
Definitely! @MartinDix, @manodeep, and I have been using the AMIP configuration for testing infrastructure changes and diagnosing issues in the UM.
>  Do we have any AMIP configuration working? 
The base branch for this PR [`dev-amip`](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-amip) is a working AMIP configuration. This is the one we'd been doing tests with so far. Prior to the new CABLE namelist, it also had LUC switched on using the CMIP6 file. When I copied over the new CABLE namelist I accidentally switched the LUC off

JhanSrbinovsky:
cool -  I think you need to re-request review so I can approve

JhanSrbinovsky (approved):

blimlim:
Thanks @JhanSrbinovsky!

Issue #51, opened 2025-03-03
Can't change MOM `io_layout` when outputting gyre/overturning diagnostics


dougiesquire:
Splitting [this comment](https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/41#issuecomment-2683676124) from @blimlim out into its own issue.

> I've looked into adding the `io_layout` setting to the ocean `input.nml`, to reduce the number of files for collation.
> 
> In each of the following cases:
> 
>   * `io_layout = 6,2`
> 
>   * `io_layout = 6,5`
> 
>   * `io_layout = 9,5`
>
>   The model crashes with a floating point error when writing the history, with the following backtrace:
> 
> 
> ```
> 1 0x00000000000b1430 nc4_convert_type()  ???:0
>  2 0x00000000000a5e57 NC4_put_vars()  ???:0
>  3 0x00000000000a5696 NC4_put_vara()  ???:0
>  4 0x00000000000290d0 nc_put_vara_double()  ???:0
>  5 0x000000000001ecf0 nf_put_vara_double_()  /scratch/tm70/tm70_ci/tmp/restricted/spack-stage/spack-stage-netcdf-fortran-4.5.2-3vzrzd2zjjdhofukp3yjzk3fyaqm2zgp/spack
> -src/fortran/nf_varaio.F90:427
>  6 0x0000000001c6a129 write_record_default()  /scratch/tm70/tm70_ci/tmp/restricted/spack-stage/spack-stage-access-fms-git.dev_2024.12.0_2024.12.0-gcn24dimfzb5xy4io5o
> duwqkt7zkulsy/spack-src/mpp/include/mpp_write_2Ddecomp.h:143
>  7 0x0000000001c6a129 mpp_write_r4d()  /scratch/tm70/tm70_ci/tmp/restricted/spack-stage/spack-stage-access-fms-git.dev_2024.12.0_2024.12.0-gcn24dimfzb5xy4io5oduwqkt7
> zkulsy/spack-src/mpp/include/mpp_write.h:27
>  8 0x000000000185f83a diag_util_mod_mp_diag_data_out_()  /scratch/tm70/tm70_ci/tmp/restricted/spack-stage/spack-stage-access-fms-git.dev_2024.12.0_2024.12.0-gcn24dim
> fzb5xy4io5oduwqkt7zkulsy/spack-src/diag_manager/diag_util.F90:2436
[...]

Issue #41, opened 2025-02-25
Collation and post processing failures


blimlim:
The collation step intermittently fails, leaving parts of the output that cannot analysed unless the failed outputs are manually collated, and corresponding UM outputs converted to netCDF. This can be time consuming, and can occur regularly when doing long simulations.

I'll start with a catalogue of the errors that have been noticed during the CSIRO runs of ESM1.6.

1. `/scratch/p66/jxs599/access-esm/archive/dev-base1-expt-7db342ca/`
```
/scratch/p66/jxs599/access-esm/archive/dev-base1-expt-7db342ca/output090 -    Exit Status:        255
/scratch/p66/jxs599/access-esm/archive/dev-base1-expt-7db342ca/output091 -   Broken pipe error
/scratch/p66/jxs599/access-esm/archive/dev-base1-expt-7db342ca/output092 -    Broken pipe error
/scratch/p66/jxs599/access-esm/archive/dev-base1-expt-7db342ca/output096 -    Exit Status:        255
```

2. `/scratch/p66/jxs599/access-esm/archive/feat-sublimation-expt-32122cc3/`
```
/scratch/p66/jxs599/access-esm/archive/feat-sublimation-expt-32122cc3/output025 - Exit Status: 255
/scratch/p66/jxs599/access-esm/archive/feat-sublimation-expt-32122cc3/output026 - Broken pipe error
/scratch/p66/jxs599/access-esm/archive/feat-sublimation-expt-32122cc3/output055 - Exit Status: 255
/scratch/p66/jxs599/access-esm/archive/feat-sublimation-expt-32122cc3/output060 - Exit Status: 255
```

3. `/scratch/p66/jxs599/access-esm/archive/feat-sand-clay-silt-expt-1fa5d2ce`
```
/scratch/p66/jxs599/access-esm/archive/feat-sand-clay-silt-expt-1fa5d2ce/output027 - Exit Status: 255
/scratch/p66/jxs599/access-esm/archive/feat-sand-clay-silt-expt-1fa5d2ce/output028 - Broken pipe error
/scratch/p66/jxs599/access-esm/archive/feat-sand-clay-silt-expt-1fa5d2ce/output057 - Broken pipe error
/scratch/p66/jxs599/access-esm/archive/feat-sand-clay-silt-expt-1fa5d2ce/output057 - Exit Status: 255
[...]

PR #55, opened 2025-03-17
Remove ocean output collation

blimlim:
Work related to #41.

Collation failures form the majority of post-processing problems which have recently caused problems during esm.6 development. Based on discussions, the preference is to remove the collation step entirely by writing complete ocean output files during the model run. 

This pull request sets the ocean `io_layout=1,1` and switches off the collation step. Additionally, compression is enabled for the MOM output with `deflate=4` to maintain similar output sizes as before. 

Adding an `io_layout` requires us to remove  the basin specific flux variables. Including these with an `io_layout` can either cause the model to crash, or silently write wrong output. See [here](https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/51) for details.

This change impacts the model walltime, increasing by ~3-7 minutes depending on the out. See [here](https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/41#issuecomment-2712197438) and [here](https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/41#issuecomment-2722901346) for details.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -113,11 +113,7 @@ submodels:
 
 # Collation
 collate:
-    exe: mppnccombine.spack
-    restart: true
-    mem: 4GB
-    walltime: 1:30:00
-    mpi: false
+    enable: false
 
 restart: /g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2025.02.17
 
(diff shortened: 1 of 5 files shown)


blimlim:
Thanks @aidanheerdegen! Merging now. I'll look into fixing the failing collation related qa test

PR #56, opened 2025-03-17
Add checksum_slow to default repro checks

dougiesquire:
As we discussed in person, this PR reconfigures the CI to:
- run `reproducibility` checksum and checksum_slow marks by default (meaning historical, restart and repeat repro tests are run)
- use payu dev
- use the latest `model-config-tests`

These settings should be revisited when we are ready to release a config.

### DIFF
--- a/config/ci.json
+++ b/config/ci.json
@@ -7,7 +7,7 @@
     },
     "reproducibility": {
         "default": {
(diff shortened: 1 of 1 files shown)


blimlim (approved):
Thanks @dougiesquire, looks good to me!

PR #62, opened 2025-03-19
Update config.yaml to point to relevant prerelease

Whyborn:
Prerelease build has additional UM flags to initialise scalars and arrays with a given value to get around non-determinism issue. Repro should fail here- if it doesn't indicates there's an issue with the tests.


Whyborn:
!test repro

PR #67, opened 2025-03-24
Checking for AMIP repro with oneAPI exe (draft)

manodeep:
Do not merge

Not really sure why the [other PR got closed](https://github.com/ACCESS-NRI/access-esm1.6-configs/pull/66) when I changed the target to `main` instead of `dev-amip`.

@blimlim I am trying to run the repro test with the exes built with oneAPI. I branched off `dev-amip` - so presumably repro should test against that - right?

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -26,7 +26,7 @@ modules:
   use:
       - /g/data/vk83/prerelease/modules
   load:
-      - access-esm1p6/pr43-7
+      - access-esm1p6/pr67-2
 
 # Model configuration
 model: access-esm1.6


blimlim:
Unfortunately automatic repro testing isn't yet working with the AMIP configurations. The existing tests use the MOM5 logs, which aren't available in the AMIP config. I have a 75% finished update for the model-config-tests to fix this. Will let you know when that's finished!

manodeep:
@blimlim Ahh okay - thanks!

Presumably I have to create another branch of the dev-PI to repro-test the PI configs - right?

blimlim:
yeah I think so!

manodeep:
Work is already done

PR #68, opened 2025-03-24
Using the oneAPI built exe (draft)

manodeep:
Do not merge

Attempt to repro check with the new oneAPI built exe


manodeep:
!test repro

manodeep:
Not needed any more

PR #74, opened 2025-03-27
Add `checks: write` Permission to Generate Initial Checksums

### LINKED ISSUE #73: `generate-initial-checksums.yml`: Add `permissions.checks: write`
See https://github.com/ACCESS-NRI/access-esm1.6-configs/actions/runs/14118693715

This is due to a later called workflow requiring `checks: write` but the caller not providing it. Update the workflow so it contains the `checks: write` permission.

CodeGat:
Closes #73

## Background

A later workflow is requiring `checks: write` but we didn't give that permission initially. Now we do.

## In this PR

* Give the repro-ci job `permissions.checks: write`


jo-basevi (approved):
LGTM

Issue #82, opened 2025-04-02
Use CMIP7 solar constant for piControl


MartinDix:
See https://github.com/ACCESS-NRI/CMIP7-Input/issues/8 for calculation of the CMIP7 value.

The piControl should use 1361.62, c.f. the currently used CMIP5 value of 1365.65.

Results from 100 year test run at https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments/issues/7

PR #84, opened 2025-04-02
Use the CMIP7 pre-industrial solar constant

MartinDix:
Update `input_atm.nml` with new solar constant.

See #82 and https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments/issues/7


blimlim (approved):
Thanks @MartinDix, looks good to me!

Issue #114, opened 2025-05-08
Use CMIP7 volcanic forcing for PI control


MartinDix:
See https://github.com/ACCESS-NRI/CMIP7-Input/issues/9 for derivation of PI values

penguian:
The parent issue to says to set the `VOLCTS_val` variable in the `coupling` namelist in the file `atmosphere/input_atm.nml` to the value 138.0, which is 1e4 times the value specified in the [externally hosted dataset documentation](https://docs.google.com/document/d/1blX5kv0We1BteqWzMKs0OuhazAcAonay/edit) at "VI – Comparison between our datasets and CMIP6".

PR #117, opened 2025-05-13
Use CMIP7 PI control volcanic AOD

MartinDix:
Following #114, change PI control volcanic AOD from 0.0134 to 0.0138 (model value is scaled by 1e4).

Change is too small to have any noticeable effect on climate.


MartinDix:
!test repro commit

penguian (approved):
Approved

PR #125, opened 2025-05-28
Update field_table (dissdet)

pearseb:
Increased the amount of CaCO3 that is dissolved when detritus is remineralised


pearseb:
Thanks Spencer!

JhanSrbinovsky:
how can I pull this Spencer? I actually already have the change so I am expecting a conflict

PR #132, opened 2025-06-09
Update field_table

pearseb:
Included a change in the field_table to parameter "disscal", which has been changed from 0.25 to 0.15. This is to reduce the dissolution of CaCO3 in the deep ocean and allow more alkalinity to make its way to the sediments via CaCO3 and be permanently buried. This is to correct having too much alkalinity in the deep ocean between 2000-3000 metres.


blimlim:
!test repro commit

PR #169, opened 2025-07-29
cice error change test

blimlim:
Pr for repro tests. See https://github.com/ACCESS-NRI/cice5/issues/60 for background


blimlim:
!test repro

PR #186, opened 2025-08-20
Testing historical reproducibility

dougiesquire:
This PR is to confirm that the changes in https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/122 do not change answers. However, in the process of setting this up noticed that some of the restart files have been modified in place (this shouldn't happen). So I'll first test whether this modification changes answers.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -28,9 +28,9 @@ platform:
 # Modules for loading model executables
 modules:
   use:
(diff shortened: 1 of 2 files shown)


dougiesquire:
!test repro

dougiesquire:
Closing as has served its purpose

PR #196, opened 2025-09-01
AMIP updates for Australian PFTs

ccarouge:
Updated config for the Australian PFTs with updated LUH states for land-cover change


ccarouge:
From Rachel:

> @ccarouge - finally got back to this today and the amip tests I'd set-up crashed with a 'UM Error - Wrong no. of atmos prognostic fields'. One thing I noticed is that there is no 'um.res.yaml' file in /g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/amip/restart/2025.09.01/atmosphere/. Could this be the problem?

MartinDix:
AMIP doesn't use the coupling fields that are present in the UM restart and fails if they're present. ESM1.6 doesn't include the reconfiguration which cleans up this sort of thing for CM2 and AM3/CM3 so separate restart files are required for AMIP and coupled.

Details on removing the extra fields to follow

rml599gh:
@MartinDix , @ccarouge - I assume this post from Spencer, https://forum.access-hive.org.au/t/access-esm1-6-development/3114/198 is the one that includes the command to remove fields. I'd forgotten this step when creating the amip restarts from the control restarts.

rml599gh:
@ccarouge - I've run some tests. Clone of the default config fails because the exe manifest doesn't match. If I set that to false I am getting an 'Atmosphere basis time mismatch'. With the restart file you created, did you reset the date to 1978 (also part of Spencer's instructions)?
I had also made a restart file myself since we wanted to test one which already had the Australian pfts somewhat spun-up. After a couple of false starts, this looks like it is running OK.

ccarouge:
@rml599gh , yes I did change the date for the restart so not sure what is happening. I'll retry tomorrow unless you prefer to use the restart you created anyway. In that case, I'm happy to get it to the pre-release area for the input files.

rml599gh:
@ccarouge - happy for you to use my restart. It is here: /g/data/p66/rml599/amip-test/restart-files/restart_Auspft. In the end, I think we'll want a restart taken from a historical run.

config.yaml
```diff
@@ -111,6 +111,9 @@ manifest:
 stacksize: unlimited
 
 qsub_flags: -W umask=027
+mpi:
+  flags:
+    - --mca mpi_yield_when_idle true
```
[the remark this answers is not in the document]
blimlim: Ah ok! Would you recommend including it or removing it here?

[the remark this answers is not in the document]
blimlim: Thanks! That sounds like a good idea, I'll remove it for the moment. Testing it out shouldn't be too urgent!

atmosphere/input_atm.nml
```diff
@@ -0,0 +1,3 @@
+&coupling
+ ocean_albedo_factor=0.92
```
blimlim, line 2: I have confirmed that this is properly read in the uncoupled configuration. Printing out the ocean albedo factor from the UM code, it used the correct value:
```
SPENCER ocean albedo factor:   0.920000000000000
```

atmosphere/um_env.yaml
```diff
@@ -5,7 +5,7 @@ DATAW: '{work_path}'
 FASTRUN: 'true'
 IDEALISE: ''
 PAREXE: ''
-PRINT_STATUS: PrStatus_Oper
+PRINT_STATUS: PrStatus_Diag
```
blimlim, line 8: For consistency with `dev-preindustrial+concentrations`

blimlim:
@MartinDix, @ccarouge, I've added some updates to sync this with the latest `dev-preindustrial+concentrations` configuration. In particular:
- Cherry picked 0b45f8a which reordered the `prefix.PRESMA` file in the `dev-preindustrial+concentrations` branch.
- Updated to the latest release exe b540461
- Added the 0.92 ocean albedo scaling. This isn't yet in the `dev-preindustrial+concentrations` branch, but I'm guessing do we want it here? 22e2526
- Cosmetic changes to match the formatting in `dev-preindustrial+concentrations`, use the same diagnostic profile, and add update to (unused) cmip7 base GHG concentrations 6804ba2, 4323bbd, 7ecba3f.
- Update the restart. I took the restart from the  [test-PIspinup-Sept-A](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/test-PIspinup-Sept-A) branch and applied the processing [here](https://forum.access-hive.org.au/t/create-an-amip-restart-from-a-coupled-restart/5261). Let me know there would be a better restart to use!


See [here for a diff compared to the dev-preindustrial+concentrations branch](https://github.com/ACCESS-NRI/access-esm1.6-configs/compare/dev-preindustrial+concentrations..191-amip-updates-for-Aust-PFTs)

Let me know if you have any suggestions!

JhanSrbinovsky:
@blimlim - my AMIP runs incorporate those updates you mentioned (well the ones relevant to CABLE anyway). I also have inplace the LUC @Whyborn and I discussed this morning via email - this was the missing piece of the puzzle it seems. I was away for a bit but it looks as though my AMIP config using this file has progressed to year 2. Once the file is in place in vk83 I'll clone a payu run using this config to verify. Note: It always did the first year happily. It was this first year LUC that was the obstacle.

Whyborn:
New LUC dataset is being added to the `modern/amip` inputs directory. When this is done, I'll update `config.yaml` and `scripts/update_landuse_driver.sh` with the relevant names.

JhanSrbinovsky:
odd. Ive never seen this happen to someone at NRI and NOT me. Issue #287 might fix this for all of us :)

blimlim:
!test repro commit

blimlim:
Just resolving the remaining conversations now, and then I think it will need a final review

blimlim:
!test repro

atmosphere/namelists
```diff
@@ -649,10 +649,10 @@ NICE=1,
  L_SEC_VAR=.FALSE.,
  L_CLIM_AERO_HGT=.FALSE.,
  AERO_BL_LEVELS=12,
- CO2_MMR= 4.314801e-04,
+ CO2_MMR= 4.3189e-04,
  O2MMR=    0.2314,
- N2OMMR= 4.143596e-07,
- CH4MMR= 4.470459e-07,
+ N2OMMR= 4.1256e-07,
+ CH4MMR= 4.4228e-07,
```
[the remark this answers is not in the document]
blimlim: I've added these in bea6696

ccarouge:
I don't think I'm qualified to do an approving review (for example I have no idea about all the STASH changes). Some comments from me:

- can we change metadata.yaml, line 5 to: ACCESS-ESM1.5 global atmosphere-land model configuration running in co2 concentration driven mode under pre-industrial forcings, as described in Ziehn et al. (2020), https://doi.org/10.1071/ES19035.
- It says "pre-industrial forcings" in the line above and underneath it, but the AMIP forcings are different, correct? Is that the historical forcings that are used?

blimlim:
Thanks @ccarouge, I've updated the metadata.yaml with those changes. I've also swapped "ACCESS-ESM1.5" -> "ACCESS-ESM1.6"

atmosphere/um_env.yaml
Whyborn: If we do plan on updating this to `PrStatus_Normal`, then we should probably do it now? I suspect the IO load will have a meaningful influence on @manodeep's optimisation work

blimlim: Added in [d2e3e2d](https://github.com/ACCESS-NRI/access-esm1.6-configs/pull/196/commits/d2e3e2dbe909737a717ac0c8fac15b1de199b021). I'll make an issue to add this change to the other configurations

Issue #207, opened 2025-09-19
Howto: add/edit the ESM1.6 configuration documentation


chrisb13:
We have an empty space now for ESM1.6 configuration documentation (not the [final url](https://github.com/ACCESS-NRI/access-esm1.6-configs/pull/190#issuecomment-3235442530)):
https://access-nri.github.io/access-esm1.6-configs/

So there's (at most) two steps to adding content, we describe two different methods, the first method only allows one to edit the landing page. 

## Getting involved in contributing

 - create sub-issues of this issue to sub-divide creation of content (`create sub-issue` button below this post) -- give issue a title of the section/pages you are working on;
 - label sub-issue with "documentation";
 - people self-assign into a sub-issue to do the task;
 - author of content nominates reviewers upon completion.

## Method 1 (takes one step but only allows one to edit existing pages)
If you just want to edit an existing page, then go [here](https://access-nri.github.io/access-esm1.6-configs/), click on the pencil twice and you can edit that page. Then follow the steps [here](https://access-nri.github.io/access-esm1.6-configs/contributing/#quick-contributions).

## Method 2 (takes two steps but you can add new pages)
If you want add a new page, then you want to add another markdown file in the folder:
https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/main/documentation/docs/pages

And then also tell the doc system that it needs to include the new page by adding a line here:
https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/e5da07046dee2e644e79fdf170dc76b2595ecca1/documentation/mkdocs.yml#L96-L98

Note you'll need to add a `pages` to the path, you can see examples [here](https://github.com/ACCESS-NRI/access-om3-configs/blob/119a52e9fcdc70c61dbb4ced47bb5a0a53f2a5e1/documentation/mkdocs.yml#L108-L126).

In the above workflows, we suggest making commits to a branch and then opening a pull request. Once a pull request is open it is possible to view a website-rendered version of the whole new website through a "PR-preview", a link will be given by the GitHub Actions inside the pull request. Typically this is of the form: 
https://access-nri.github.io/access-esm1.6-configs/pr-preview-#PR-number

Once changes are merged into the main branch, it should update [the build of the docs here](https://access-nri.github.io/access-esm1.6-configs/).

Feel free to ask questions in this thread if any aspect of this workflow is unclear. 

--

[...]

PR #208, opened 2025-09-19
Minor config change to esm1.6 config docs

chrisb13:
Fixing a broken link and turning off stubs

Noticed the above two, when writing [these instructions](https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/207).


anton-seaice (commented):
I think just delete the whole plugin sections. Commented out code is normally just more trouble than its worth (unless there a clear purpose / relevance for it)

anton-seaice (approved):
Looks good

PR #216, opened 2025-09-24
re-order field records

JhanSrbinovsky:
What I actually want to do here is modify titles, long names for some STASHfields, so that they more accurately and consistently describe the actual field. Once started down this path there were a dozen things present that I have neglected to revise since CMIP5 days. Re-ordering, field names, option codes, etc. I got 90% of the way through this and realized it was going to be a pain to review. So I am going to do this in a series of steps. Hopefully we can standardize this STASHmaster across CMIP7 models, or at least get close. There is no hurry to get this into a September SpinUp.

Here as a first step, the file is recorded  - there are zero changes to the "code" here. Just shuffling the order and neatening up the file.

Although perhaps  unnecessary, I ran 2 back-back jobs:

`git clone git@github.com:ACCESS-NRI/access-esm1.6-configs.git -b ReorderSTASHmasterFields ReorderSTASHmasterFields-2
`
And a control run:

`git clone git@github.com:ACCESS-NRI/access-esm1.6-configs.git -b dev-preindustrial+concentrations dev-preindustrial+concentrations-24Sept`

where `dev-preindustrial+concentrations` is the base of `ReorderSTASHmasterFields`.

Comparing output data fields from, monthly means at the end of Jan, between netcdf files, produced by both runs, with nccmp would shows that the 2 are equivalent as expected

```
2025-09-24 11:38:38.783345 +1000 INFO ../../src/nccmp_data.c:3818 Thread 0 comparing data for variable fld_s33i002 at record 0.
2025-09-24 11:38:40.261289 +1000 INFO ../../src/nccmp_data.c:4093 Finished comparing group data.
2025-09-24 11:38:40.261707 +1000 INFO ../../src/nccmp_group.c:135 Group comparison complete.
2025-09-24 11:38:40.261858 +1000 INFO ../../src/nccmp.c:109 Comparisons complete.

```


blimlim (approved):
Thanks @JhanSrbinovsky for breaking the changes up into smaller PRs, it's definitely helpful for reviewing!

These changes look good to me. I've double checked that each entry in the updated file corresponds exactly to an entry in the original version – all looks good. 

Thanks for running the extra test too

JhanSrbinovsky:
@blimlim Once I've converted a PR to a draft, can I convert it back again?

PR #217, opened 2025-09-24
Edit option codes for (CABLE&CASA) STASHmaster records

JhanSrbinovsky:
based on ReorderSTASHmasterFields-2 branch #216

From [UM documentation papaer](https://code.metoffice.gov.uk/doc/um/latest/papers/umdp_C04.pdf)

From the section defining the 20th digit of the option code:

n20 Thermal Vegetation Canopy Indicator, and sea-surface and sea-ice indicator.

0 Item not dependent on thermal vegetation canopy.
1 Item included for snow canopy scheme.
2 Item included if sea-ice meltponds are used in surface albedo calculation.
3 Item included if sea-surface freezing temperature is salinity-dependent.
4 Item included if prognostic sea ice roughness length is in use.

Currently “2” is selected. Clearly t his is not appropriate. Replace with “0”.

The next 20 digit code represents the version mask. The documentation around this is very hard to decipher. As it is doesn’t seem to be causing any problems I’ll leave it alone for now.


JhanSrbinovsky:
this is tested now as well and similarly first monthly mean outputs are identical.

MartinDix:
The version masks for some of the CABLE diagostics are strange. E.g.
```
1|    1 |    3 |  801 |SOIL LAYER 1 TEMPERATURE ON TILES   |
2|    0 |    0 |    1 |    2 |    5 |   -1 |   -1 |    9 |    1 |    9 |    0 |
3| 000000000000000000000000000000 | 00000000000010000000 |    3 |
```
and probably should be consistent with the prognostics. However as long as they're actually gettting saved it doesn't really matter.

JhanSrbinovsky:
> The version masks for some of the CABLE diagostics are strange. E.g.
> and probably should be consistent with the prognostics. However as long as they're actually gettting saved it doesn't really matter.

I can't find where I tried to change the VersionMask to something which I thought made more sense and I definitely think I tried to change it to the same as the prognostics. Initially I tried to copy the prognostic records and just change the Section Number, but it wouldn't work. I progressively ended up changing several fields in the record to accommodate. I remember the version mask issue and I concluded that it musn't mean what I thought it did, or be constructed that way

JhanSrbinovsky:
Ive been looking into what I did with the "Version Mask". On the ESM1.6 side I initially tried to just copy the prognostic records(section 0) to the diagnostic section(3). This didn't work and code by code I went through and changed it to what made sense/what it was. This is where I [changed the Version mask FROM what it was in the prognostics and back to what it was in the diagnostics](https://github.com/ACCESS-NRI/access-esm1.6-configs/commit/4451d8fd9645047af67ee5259280c622fbb8f4bc). This also seems to be what the other UM fields are doing. It reinforced for me that I must not know what this Version Mask is doing. Reading the documentation didn't clarify things for me. Given that this was consistent with how existing UM record were, and it worked, I gave up on trying to fix (or make sense of) what wasn't broken.

blimlim:
> Let me know how you would like too proceed?

Happy to go with whichever way is simplest on your end! 

Should I review this one as is, or will there be a bit of reshuffling?

blimlim:
!test repro

blimlim (approved):
Thanks @JhanSrbinovsky, the changes look good to me!

blimlim:
With the name changes, e.g.
```diff
- 1| 1 | 0 | 801 |SOIL LAYER 1 TEMPERATURE ON TILES |
+ 1 | 0 | 801 |CABLE SOIL TEMPERATURE LAYER 1 (K) |
```

should these new names also go into the converted netCDF files? If so, I can add a PR to the [conversion script](https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/src/um2nc/stashmasters.py) to match the changes

JhanSrbinovsky:
> should these new names also go into the converted netCDF files? If so, I can add a PR to the [conversion script](https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/src/um2nc/stashmasters.py) to match the changes

That would be great. I was going to ask about how I can do that but I want to update the AM3 version first so the I can add in the Urban tiles stuff for Jiachen

PR #218, opened 2025-09-25
field record description or long names edited to match AM3

JhanSrbinovsky:
Edit title description of CABLE records to be consistent and match AM3


JhanSrbinovsky:
again back-to-back pay runs for a month, nccmp shows output to be identical @blimlim 

both pay clones used dev-PI+concentations. git log shows commits were made - I suspect payu forces to git to make them, but to where, just to local clone? There's no way of referencing the rest jobs is there?

JhanSrbinovsky:
@blimlim - reminder that this is low priority to get into the September SpinUp so don't stress. From here on all these PRs are based on the preceding branch. NOT dev-PI-concentrations. I can deal with the conflicts but I was advised to do it this way. In the past with svn I would just branch them all off  dev-PI-concentrations and keep rebasing  each PR-branch. if there is a better way with GIT then Im happy to do this. Rebranching and merging these changes in is not a big problem.

JhanSrbinovsky:
reopen

PR #219, opened 2025-09-25
rm-ed redundant PP number

JhanSrbinovsky:
Redundant and potentially misleading PPFC number (very likely it was me)- was originally copied from some field record 100 years ago and remains here for no apparent reason

Again from [UM docs](https://code.metoffice.gov.uk/doc/um/latest/papers/umdp_C04.pdf): PPFC number

PPFC PP field code. This is the field code defining the data to (older) PP graphics applications and is only used
to specify the PP header item (23=lbfc) on output. A list of currently reserved PP field codes is maintained
at the Met Office; if further details are required please contact the UM System Team. Example lbfc codes:
0 unspecified for PP applications
1 height
8 pressure
16 temperature

The number we had in there was 1510. IDevenK where this comes from!


MartinDix:
Likely originally copied from `SURFACE TEMPERATURE ON TILES`. Not used in any of our post-processing.

PR #220, opened 2025-09-25
revised diagnostics entries to match prognostics

JhanSrbinovsky:
Here we add the diagnostic records (section=3) to match the prognostic records (section=0)

Note: STASH records 907:915 in section. 3 diagnostics don’t have a counterpart in prognostics - section 0
 
From [UM docs](https://code.metoffice.gov.uk/doc/um/latest/papers/umdp_C04.pdf): PPFC number

```
#|Model |Sectn | Item |Name |
#|Space |Point | Time | Grid |LevelT|LevelF|LevelL|PseudT|PseudF|PseudL|LevCom|
#| Option Codes | Version Mask | Halo |
#|DataT |DumpP | PC1 PC2 PC3 PC4 PC5 PC6 PC7 PC8 PC9 PCA |
#|Rotate| PPFC | USER | LBVC | BLEV | TLEV |RBLEVV| CFLL | CFFF |

```
The “Space” code in section 0 variables (=3) is appropriate for prognostics. (See belo) exert from UM doc paper). These were edited =0 for diagnostics appropriate for prognostics.

Space Code
```
The space code. Specifies the space requirements of a STASH item. If a diagnostic is calculated only
when required by STASH (i.e. under a STASH flag), or is only copied to STASHwork under a STASH flag,
then the space code is 0. The possible values of space code are:
0 Diagnostic field for which space is required only when the diagnostic is requested.
2 Section 0, 33, 34 or 54 items only: primary field available to STASH.
3 Section 0, 33 or 34 items only: primary field unavailable to STASH which is addressed in the dump
and D1. etc etc
```

Otherwise, the records match the prognostic side records.


JhanSrbinovsky:
Attached are two files for your convenience. They are both subsets cut out of the STASHmaster file. prognostics/diagnostics

[diag-220.txt](https://github.com/user-attachments/files/22528022/diag-220.txt)
[prog-220.txt](https://github.com/user-attachments/files/22528024/prog-220.txt)

JhanSrbinovsky:
There is something amiss here. The nccmp of data does not entirely check out. Investigating now

JhanSrbinovsky:
Each record has 5 lines. 
```
#|Model |Sectn | Item |Name |
#|Space |Point | Time | Grid |LevelT|LevelF|LevelL|PseudT|PseudF|PseudL|LevCom|
#| Option Codes | Version Mask | Halo |
#|DataT |DumpP | PC1 PC2 PC3 PC4 PC5 PC6 PC7 PC8 PC9 PCA |
#|Rotate| PPFC | USER | LBVC | BLEV | TLEV |RBLEVV| CFLL | CFFF |
```

The 1st line containing the name change is fine.

The 2nd line we have replaced grid=2 with grid=21. This results in differences exposed by nccmp. This was not expected given the  description of grid=2and21. I'll investigate this further AFTER submitting a revised file that does pass nccmp testing.

The 3rd line - unexpectedly the version Code seems to be causing a problem. It was always expected that diagnostics (section=3) records would expose different treatment of the record's values in general BUT this causing no difference in the prognostic records I didn't think the version code would be a problem here. nonetheless it is and has been reverted to its previous value

The 4th line change is fine.

The 5th line change is fine.

In summary, lines 2 and 3 which had no effect on the prognostics were reverted when those same changes were  imposed on the diagnostics in order to maintain bitwise reproducibility. 

It may be neater then to revert these same lines in the prognostics such that the records are the same? OTOH this does represent a point of difference between/n the two - although it isn't a point of difference that has any impact on the prognostics section.

atmosphere/prefix.PRESM_A
```diff
@@ -1028,581 +1028,580 @@
 4|    1 |    2 | -99  -99  -99  -99  -99  -99  -99  -99  -99  -99 |
 5|    0 | 1563 |    0 |  129 |    0 |    0 |    0 |    0 |    0 |
 #
-1|    1 |    3 |  801 |SOIL LAYER 1 TEMPERATURE ON TILES   |
-2|    0 |    0 |    1 |    2 |    5 |   -1 |   -1 |    9 |    1 |    9 |    0 |
+1|    1 |    3 |  801 |CABLE SOIL TEMPERATURE LAYER 1 (K)  |
```
JhanSrbinovsky, line 1031: these are all of the diagnostics being brought into line with prognostics

PR #205, opened 2025-09-17
compressed lice input file

anton-seaice:
This version of the pesudo iceberg spreading mask:

- is compressed using netcdf4 compression
- only includes the needed variable (FICEBERG_GC3)
- and has been pruned to 7 sig figures

this results in a reduction from 40MB to 1MB in file size. The pruning in sig figures means that it will be slightly answer changing

contributes to #204

PR #221, opened 2025-09-25
Compressed lice input file

anton-seaice:
Cheery pick from #205 

Should we do this @blimlim - or do on mass with other changes next week?


blimlim:
> Should we do this @blimlim - or do on mass with other changes next week? 

Let's wait for the new PI config and then sync the changes together!

PR #228, opened 2025-10-01
test-PIspinup-Sept-C

JhanSrbinovsky:
@ofa001 has found that 

The reason for starting earlier in the run is that  by year 600 of AprilSpinUp (where Sept-A begins), 500 after the solar constant change, the region in the Southern Ocean has already cooled down considerably and is already spilling water into the excessively cold water of deep ocean in the rest of the model (ocean timescale to reverse this is long).

This suggests we restart the model from an even earlier time. However there is a significant step change in ocean dynamics due to a revision to ocean BGC at approximately year  280(experiment counter=382). This suggests we start after this has occurred. As such the decision was made to start from restart300(experiment counter=402), 200 years earlier that Sept-A. 

Technically:
--------------
Only modifications to these two files are  relevant

```
  atmosphere/input_atm.nml - includes revised ocean albedo scaling factor (0.92)
  config.yaml                             - includes path to new restarts in config
```
  
  I didnt want to add these binaries to the repo.

  Can you please put them in vk83/ and create another branch test-PIspinup-Sept-C.

### DIFF
--- a/atmosphere/input_atm.nml
+++ b/atmosphere/input_atm.nml
@@ -7,4 +7,5 @@
  ocn_sss=.false.
  sdump_enable=.false.
  rdump_enable=.false.
+ ocean_albedo_factor=0.92
 &end
--- a/config.yaml
+++ b/config.yaml
@@ -16,7 +16,7 @@
 # Note: if laboratory is relative path, it is relative to shortpath/$USER
 laboratory: access-esm
 
-jobname: pre-industrial
+jobname: test-PIspinup-Sept-C
 queue: normalsr
 walltime: 2:30:00
 jobfs: 1500MB
@@ -119,7 +119,7 @@ submodels:
 collate:
     enable: false
 
-restart: /g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2025.09.26
+restart: /g/data/p66/jxs599/ESM16/PAYU/Feat/feat-Sept-C/restart300 
 
 # Timing controls
 calendar:
--- a/manifests/restart.yaml
+++ b/manifests/restart.yaml
@@ -2,138 +2,138 @@ format: yamanifest
 version: 1.0
 ---
 work/atmosphere/restart_dump.astart:
-  fullpath: /g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2025.09.26/atmosphere/restart_dump.astart
+  fullpath: /g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2025.09.26.B/atmosphere/restart_dump.astart
   hashes:
-    binhash: eeebd0571fe00ea95219c1271c181992
-    md5: d1c33af7ee7609f049c4ad2ab37adbe0
+    binhash: 81c75f027ed1fdd9e2a34d673cf58986
+    md5: e013ca06e1d0acc356967e9f72d5a79c
 work/atmosphere/um.res.yaml:
-  fullpath: /g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2025.09.26/atmosphere/um.res.yaml
+  fullpath: /g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2025.09.26.B/atmosphere/um.res.yaml
   hashes:
-    binhash: b79747f8d84be69b6264ed0a559bee99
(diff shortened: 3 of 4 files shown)


JhanSrbinovsky:
thanks @blimlim - Weird - it's running OK. I don't remember changing it before, maybe a CICE5 thing?. If you could change it that would be great :)

 Anyway, in my travels I noticed an amendment you might want to make in 

`/g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2025.09.26.B/README`

it is missing that ocean/ comes from restart990  AT

`/g/data/p73/archive/CMIP7/ACCESS-ESM1-6/spinup/April2025/restart990/ 
`
quoted path is where it is copied to not from

JhanSrbinovsky:
I mean like Sept-A README says 602 came from AprilSpinUp restart500, which is handy

Its pretty confusing the offset of 101 corresponding to restart000 

SeptB 1092 doesn't say this the same way.

I think initially it came from AprilSpinUp restart990. At least the wombat restart checks out bitwise. Perhaps the JuneSpinUp was configured from here, I can't remember.

ofa001:
Hi @JhanSrbinovsky, sorry just caught up with this message chain today, didn't see any messages in my inbox yesterday, though I was unwell so they may have slipped past when I looked in the evening.

The reason for starting earlier in the run is buy year 600 of AprilSpinUp,  500 after the solar constant change  the region in the Southern Ocean has already cooled down considerably and is already spilling water into the ecessively cold water into deep ocean of the rest of the model (ocean timescale to reverse this is long).  

My preference is to restart the run as earlier as possible in AprilSpinUp as discussed with @MartinDix and Rachel. 

The compromise however was to do it after the BGC change had gone in, is so that's why a restart in the ocean/atmoshere/BGC set up around year @400 came in so 200 years earlier than Sept-A.  For the  ice restarts a similar field as used for CICE5 as the one provide by @anton-seaice for the Sept-A/B runs. But as you say there is an issue with Leap years.  I will have a look at the test run you have done.

blimlim:
Hi @JhanSrbinovsky, because the two restarts are 200 years apart, the leap year mismatch will occur every 400 years - the components won't agree on the run length which can lead to crashes.

> Should we make a branch for test-PIspinupSept-C and use this as the restart? Or just leave it as is?

If you prefer to keep going with the current run, we'll need to pause it and fix the ice restart. To pause the run, you can create an empty file named `stop_run` in the control directory. This will prevent the next resubmission from going ahead, and so the restart can be fixed up in-between. There's some more detail on this [here](https://payu.readthedocs.io/en/stable/usage.html#:~:text=If%20you%20have%20instructed%20payu%20to%20run%20for%20a%20number%20of%20resubmits%20but%20for%20some%20reason%20need%20to%20stop%20a%20run%20after%20the%20current%20run%20has%20completed%20create%20a%20file%20called%20stop_run%20in%20the%20control%20directory.)

The change to the restart won't be obviously recorded in the runlogs and experiment archive though, and the fixed up restart could get automatically deleted by payu,  and so for documenting/tracking the experiment it might be better to start from a new branch.

ofa001:
The other issue @blimlim and @JhanSrbinovsky is with the restart is it having the wrong timestamp, which would be corrected if we totally restart, is that the ice output files start from the year 602 not 402, which if we ever tried to difference cross compare with Sept-A run could cause confusion down the line. The actual ice fields in the history files are  sensible in the run. 
The ice model will adjust to new SST conditions from the new ocean model restart file relatively quickly, ('decadal timescale at most')

JhanSrbinovsky:
Thanks @ofa001 for that explanation. 

ok @blimlim - looks like we need a -C version of test-PIspinup-Sept-A. I've got payu sync change locally, and jobmname changes but these dont need to be pushed. The important changes are to re=locate the restarts which you have already done and change the ice restart date.

It looks like payu sync must've timed out as the rsync is only partially complete. I suspect this will pick it up next time anyway. Can we organize the Sept-C config soon so we don't lose too much ground on this?

blimlim:
Just to make sure I understand properly, is the plan for the September C config to match this branch but with the corrected restart?

blimlim:
Hi @JhanSrbinovsky, I've made a new [test-PIspinup-Sept-C](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/test-PIspinup-Sept-C) branch you can use. This matches test-PIspinup-Sept-A, but swaps to the septC restart copied to vk83. This has the corrected ice date too

JhanSrbinovsky:
Excellent, thanks @blimlim

ccarouge:
@JhanSrbinovsky Mind if I remove the review request for now because I don't think we are intending to merge this now.

Issue #238, opened 2025-10-21
Update JULES PFT parameters in namelists[RUN_PFT]


JhanSrbinovsky:
Back when we first introduced the Australian PFTs as simply copies of 2 & 4, we were expecting results to be the same. We  then realized there would be slight differences due to computational numerics that could be expected to become significant differences in weather. However, comparing per tilmestep evolution of these differences still didn’t add up. 

It turned out that this was due to JULES albedo parameters ( in `atmosphere/namelists[RUN_PFT]`)on the first tilmestep being used as we dodged albedo on the first tilmestep. The last two columns (reflecting PFTs 12 &13) were zero. Once changed to reflect 2&4, things made more sense. (Considering the albedo situation).

In the first instance these changes have not permeated to the official config, which we introduce here.

Analysis of Atmospheric optical depth due to dust has revealed that there is zero dust being contributed by the Australian continent, where as prior to the introduction Aust PFTs there was. This could be due to the zero value given to roughness parameters in `atmosphere/namelists[RUN_PFT]`.
  

This has been corrected locally for OctA,B&C runs. I have noted the year(s) where this update kicks in under Further Notes Section in the [experiments log](https://forum.access-hive.org.au/t/esm1-6-spin-up-experiments/4219)

PR #239, opened 2025-10-21
238 update jules pft parameters in namelistsrun pft

JhanSrbinovsky:
see #238 

Note: The appropriateness of parameter values for EGBL and DBL @har917 @rml599gh @NortonAlex


ccarouge:
I've just seen that now. Do we still want it in since we determined it has no effect on the dust? Do we only want to change the albedo parameters since the other ones don't seem to have an effect or is it better to change them all? @JhanSrbinovsky @har917

ccarouge (approved):
LGTM

PR #245, opened 2025-10-27
corresponding cable.nml change to removal of cable_runtime%esm15 in IF condition

JhanSrbinovsky:
https://github.com/ACCESS-NRI/UM7/pull/173

reqd switch to ensure same behaviour as that previously enabled by now (semi-)defunct esm15 switch. This switch has always been in this conditional, just not activated as esm15 switch was


ccarouge (approved):
Fine by me. Makes things clearer for users.

PR #252, opened 2025-10-30
Update mkdocs.yml to fix link problem

### LINKED ISSUE #250: Top level docs links lead to 404
The top level links in the `ACCESS-ESM1.6 Documentation Overview` section lead to 404s, but the sidebar links work fine. They work in the OM3 docs, and I can't tell what was done differently here.

chrisb13:
Closes: https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/250 (if works)

As suggested by @atteggiani here: https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/250#issuecomment-3465762918


atteggiani (changes requested):

ccarouge:
@Whyborn this is now building fine for the preview. Please check it solves the issue #250

Whyborn (commented):
I think it resolves the fundamental problem, which was incorrect resolution of absolute links, but the Inputs and Configuration/Configuration choices pages still lead to 404s, just because the links in the top `index.md` are incorrect (There is no `/inputs/Forcing-data-models` nor `/configurations/Overview`). Not sure if you want to fix this here, as it's a different problem.

ccarouge:
> (There is no /inputs/Forcing-data-models nor /configurations/Overview). Not sure if you want to fix this here, as it's a different problem.

I would leave this to a PR updating the Home page.

Whyborn (dismissed):
With the caveat that the remaining broken links (which are broken due to pages actually not existing, rather than absolute links not being resolved) will be fixed in a later PR, this looks fine to me.

anton-seaice:
Could you re-review please @atteggiani ?

atteggiani (approved):
Looks good to me

PR #258, opened 2025-11-06
257 test revised pft parameters

JhanSrbinovsky:
aim is to test revised crop parameters. New executable requires new PFTlookup


config.yaml
```diff
@@ -28,9 +28,9 @@ platform:
 # Modules for loading model executables
 modules:
   use:
-      - /g/data/vk83/modules
+      - /g/data/vk83/prerelease/modules
```
JhanSrbinovsky: will replace with build of new main: However this was the testing build used in PR of latest merge to main

config.yaml
```diff
@@ -120,6 +120,7 @@ sync:
 
 
 userscripts:
+    setup: bash ./copy_nmls.sh
```
JhanSrbinovsky: Hack new PFT lookup into place

ccarouge: Why not point to the new file? Even if it is in your own directory for now.

Issue #242, opened 2025-10-22
Remove `melt` diagnostic


anton-seaice:
In MOM5, there is a `melt` diagnostic support to represent _water flux transferred with sea ice form/melt_ but it is incorrect

`melt` is calculated as the (salt flux from sea ice) divided by ( reference concentration of salt in sea ice) and is equal to the cice diagnostic `dvidtt`/`sidmassth` - which is the change the sea ice due to thermodynamics.

`sidmassth` and therefore `melt` includes melt/freeze of sea ice from process which do not interact with the ocean (flooding of snow, and evaporation/sublimation/condensation/deposition)

`melt` is not prognostic - it is only used in MOM5 if running with a data atmosphere and needing to balance global ocean mass (i.e. if `zero_net_water_coupler=.true.` is set)

We should remove the `melt` diagnostic, the correct ones are `wfimelt` and `wfiform` for _water into/out of ocean due to ice melt/form_ (based on the coupled sea ice melt and sea ice form fluxes) 

(p.s. the salt flux from sea ice <-> ocean includes salt exchange for flooding and evaporation as there is no salt fluxes with the atmosphere, and the sea ice has a constant concentration of salt)

@dougiesquire

ofa001:
@anton-seaice @dougiesquire The background to the wfimelt and wfiform diagnostics is they were set up for CMIP5 when they data request from the ocean community wanted the flux separated into the two components.   You are right though some of the terms in fresh (come from snow) could be a melt term whilst its freezing within the same grid box in a different ice category.

 The definition of melt from MOM5 has always had issues its defined as

melt =-Ice_ocean_boundary%salt_flux(i,j)*ice_salt_concentration_r

Depending which run you took the plot from Anton, hopefully after latest fixes on salt_concentration value bug  that Martin put in ESM1.6 namelist recently when we were trying to track down salt budget issue, (https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/232#issue-3511892906 ) it wasnt the issue there as it was  likely both multiplying and dividing out but here its a direct use of the value.

Melt is the salt flux whilst wfimelt wfiform is the water flux which also contain the snow effect, which is addtional difference when making a comparison.

PR #261, opened 2025-11-10
Remove melt diagnostic

anton-seaice:
wfimelt and wfiform are the correct diagnostic for waterflux due to the sea ice thermodynamics

melt does not represent a process in esm1.6 - it's only relevant when using salinity restoring in the ocean 

contributes to #242


anton-seaice:
!test repro

anton-seaice:
!test repro

anton-seaice:
!test repro

PR #264, opened 2025-11-10
Add the oa3-mct v5.2 namcouple and use the deployed exe

manodeep:
Testing for bitwise repro

Marking as draft - do not merge.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -30,7 +30,7 @@ modules:
   use:
       - /g/data/vk83/prerelease/modules
   load:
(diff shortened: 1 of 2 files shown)


manodeep:
!test repro

manodeep:
!test repro

manodeep:
These tests should have been bitwise identical - I am going to check with two separate build with my gadi-spack install and check manually.

manodeep:
Saving inodes. Will check again

PR #274, opened 2025-11-17
Q219 redundant pp rebranch

JhanSrbinovsky:
rebased #219


JhanSrbinovsky:
@blimlim I could've just rebased the original #219 branch, especially seeing as though I had already smuggled #217 in along with #218. I noticed that I left the PP number in the diagnostic records. I vaguely remember that I had tested this. I wouldn't be entirely surprised that IF it is a diagnostic, the record needs a PP number  to be specified (even if it is a meaningless number). I want to test it using the CI !testrepro was it?

JhanSrbinovsky:
!test repro

blimlim (commented):
Thanks @JhanSrbinovsky, changes are looking good to me. Just had one change for the whitespace.


I noticed some of the section 3 variables still have the 1510 pp code, e.g:
```'1|    1 |    3 |  174 |TRANSPIRATION FROM CABLE (ON TILES) |
2|    0 |    0 |    1 |    2 |    5 |   -1 |   -1 |    9 |    1 |    9 |    0 |
3| 000000000000000000000000000000 | 00000000000010000000 |    3 |
4|    1 |    2 | -99  -99  -99  -99  -99  -99  -99  -99  -99  -99 |
5|    0 | 1510 |    0 |  129 |    0 |    0 |    0 | 9999 |  339 |
```
Should these be edited as well, or left as is?

JhanSrbinovsky:
Finally, I was using Oct-B-combined as a control. I made all the changes that were in the control BUT I was still getting tiny differences in the output. Of the "seemingly" trivial remaining differences in config.yaml one is the run length.  So now it is running for a year. control/atmosphere ocean/ coupled/ ice/ are replaced by those directories in Oct-B-combined. The only remaining difference in config.yaml is the job name:. The manifests/ directory is also different b/n the jobs. Because this is taking so long  

I cloned the "q-219" config as both "q-219-control" "q-219-test" changed them both to run for 1 month. The ONLY difference b/n the two jobs is that atmosphere/prefix.PRESM_A in the  "q-219-control" job is that from the HEAD of dev-PI=concentrations prior to making any changes to it.  The version in "q-219-test" is that in the PR-ed branch

```
nccmp -df q219-control.../output000/atmosphere/netCDF/aiihca.pa-219801_mon.nc q219-test.../output000/atmosphere/netCDF/aiihca.pa-219801_mon.nc
```

So now that is settled, I'll test altering the PP field for the Section 3 records.

blimlim (approved):
Thanks @JhanSrbinovsky, changes look good to me!

I had a read of the docs to understand the `CFFF` changes, but wasn't successful. It doesn't seem to have an effect for us, so looks good on my end

Issue #80, opened 2025-04-01
Create test config for cice5 integration


anton-seaice:
Create a demo config for cice5 work, using https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/70 build and [coupling updates](https://github.com/ACCESS-NRI/cice5/compare/access-esm1.6...ACCESS-NRI:cice5:24-set-up-cice5-for-esm156)

anton-seaice:
I put a start in here: 

https://github.com/ACCESS-NRI/access-esm1.6-configs/compare/dev-preindustrial+concentrations...80-dev-preindustrial+concentrations+cice5

It needs some payu modifications to run ... I am working on those in https://github.com/payu-org/payu/issues/572

anton-seaice:
I've done some runs with this configuration now.  The configuration is based on the esm1.6 configuration from about March - similar to the PI-02 configuration. The labels are as follows:

- cice5-db: cice5 with the [parameters](https://github.com/ACCESS-NRI/access-esm1.6-configs/compare/dev-preindustrial+concentrations...80-dev-preindustrial+concentrations+cice5) similar to @DaveBi 
- pi-02: PI-02 output on gadi
- cice4 baseline to match the cice5 run from [this repo](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/47c40b25e53a78760901e33da8c230ce2a5d67b1)
- c5+bubbly: same as the other cice5 run but with 'bubbly' conductivity instead of 'MU71'

![Image](https://github.com/user-attachments/assets/315e7ad1-2e37-483d-82e1-50ff88f4fccb)

![Image](https://github.com/user-attachments/assets/aebdf0ca-0549-400d-a9b0-0ef0cd3576a4)


And compared to modern day Observations (purple):

![Image](https://github.com/user-attachments/assets/699d867e-2a7a-4635-836d-6e3e3c76aaa2)


Southern hemisphere results don't look that good. I will try setting chio / sinw and cosw as suggested by @ofa001 

I assume the initial conditions will not be impacting after a few years ... but should we be trying with some more realistic southern hemisphere initial conditions ?

ofa001:
Hi @anton-seaice,  Yes the 1m initial conditions in the SH could be impacting these results still, especially the first few summers, but I am surprised the winters the ice extent is also greater than normal  the difference between bubbly and MU71 is minor as I suggested.  

[...]

PR #426, opened 2026-02-04
Notes on CICE5 in ESM1.6

anton-seaice:
Some notes related to #80 . This is a very short summary of the CICE5 configuration for esm1.6, the main reference will be the not-yet-written paper

@ClareCR @blimlim - should we have a seperate page for coupling, and put the meltwater stuff there?


documentation/docs/pages/model_components/cice.md
```diff
@@ -0,0 +1,22 @@
...
+
+In ACCESS-ESM1.6, the sea ice component is CICE5 [@hunke2015cice] updated from CICE4 used in ACCESS-ESM1.5.
+
+Scientifically the sea ice model is configured the same as ESM1.5 [@Ziehn2020]. The scientific configuration is summarised as follows:
+
+- Zero-layer thermodynamics (Semtner 1976)
+    - One layer of snow and one layer of ice
+    - UM calculates ice surface temperature, and conductive heat flux into the sea-ice
+- Ice transport - Lipscomb (2001) and ridging – Rothrock (1975)
+- Internal Ice Stress follow EVP (Hunke and Dukowicz, 2002)
```
blimlim: Should the references here added to the bibliography?

documentation/docs/pages/model_components/cice.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Meltwater Runoff
+
+Like ESM1.5, the OASIS3-MCT coupler is used and the sea ice model acts as the interface between the atmosphere and ocean models. The only significant change to this interface since ESM1.5 is changes to meltwater from Antarctica and Greenland. As there is no ice sheet model, the volume of meltwater dischange from Antarctica and Greenland is equal to the instantaneous precipitation over each continent. In ESM1.6, this is partially discharged at the coastline of each continent (to represent ice shelf basal melt) and partially spread in open ocean (to represent melt from icebergs). In ESM1.5 all meltwater is at the coastlines. In addition, the latent heat to melt this water is now taken from the ocean. Meltwater runoff is configured in the `input_ice.nml` [namelist](https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/dev-preindustrial%2Bconcentrations/ice/input_ice.nml#L14-L25) with a presrcibed pattern from the [lice dischange](https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/13cc7d229b0d4bda193879b8b30cde3441d61bec/config.yaml#L98) input file.
```
ccarouge:
```suggestion
Like ESM1.5, the OASIS3-MCT coupler is used and the sea ice model acts as the interface between the atmosphere and ocean models. The only significant change to this interface since ESM1.5 is changes to meltwater from Antarctica and Greenland. As there is no ice sheet model, the volume of meltwater discharge from Antarctica and Greenland is equal to the instantaneous precipitation over each continent. In ESM1.6, this is partially discharged at the coastline of each continent (to represent ice shelf basal melt) and partially spread in open ocean (to represent melt from icebergs). In ESM1.5 all meltwater is at the coastlines. In addition, the latent heat to melt this water is now taken from the ocean. Meltwater runoff is configured in the `input_ice.nml` [namelist](https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/dev-preindustrial%2Bconcentrations/ice/input_ice.nml#L14-L25) with a prescribed pattern from the [ice discharge](https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/13cc7d229b0d4bda193879b8b30cde3441d61bec/config.yaml#L98) input file.
```

The discharge input file is call "lice", unless it means something specific for the ice model, I would keep "ice discharge input file" in the text (otherwise the text makes me think of nits). Or add pre-formatting or quotation marks ("). Choose what would work best in this case.

ccarouge (changes requested):
Looks good. I spotted a few typos, nothing major. I would keep the meltwater portion and decide later if it needs to move depending on what documentation gets written.

I would approve but fixing the typos is going to cancel my approval anyway, I believe.

documentation/docs/pages/model_components/cice.md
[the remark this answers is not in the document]
anton-seaice: ah - we've always said there's no time lag though, so is the routing through the river model only a timestep or two ?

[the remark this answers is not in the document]
anton-seaice: I also thought the snow wasn't removed, just an equivalent volumne was added as meltwater

anton-seaice: Thanks - i used the filename explicitly in then end to be clear.

I guess the _lice_ is liquid ice ? Not sure. Its not sea ice though, so needs some qualifying

blimlim: Maybe land ice? Not sure either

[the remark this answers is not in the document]
ofa001: It would be good to get @DaveBi to comment on what you write.

Issue #224, opened 2025-09-30
Diagnostic file naming scheme


aidanheerdegen:
I am currently [working on methods to add appropriate metadata to diagnostic output files](https://github.com/ACCESS-NRI/addmeta/pull/66) as a [post-processing step in payu](https://github.com/payu-org/payu/issues/577).

It appears ACCESS-OM3 and ACCESS-ESM1.6 are diverging on naming schemes. ACCESS-OM3 has [decided on a naming scheme](https://github.com/ACCESS-NRI/access-om3-configs/issues/374#issuecomment-2750096126) that is different to ACCESS-ESM1.6 (which as far as I can tell hasn't documented this, but the [MOM naming scheme was inherited from COSIMA](https://github.com/COSIMA/access-om2/issues/185#issuecomment-600994799)).

For the purposes of the project to add metadata to diagnostic output files it isn't necessary that the file naming scheme be identical, but it would be good if they at least shared a common controlled vocabulary. They don't appear to currently, e.g.

| Field | ESM1.6 | ACCESS-OM3 |
|--------|--------|--------|
| frequency | `hourly`, `daily`, `monthly`, `yearly`, `snap` | `1hr`, `day`, `mon`, `yr`, `fx` |
| averaging method | `mean`, `min`, `max`, `snap`, `rms`, `pow02` | `mean`, `snap` ??? |

See also https://github.com/ACCESS-NRI/access-om3-configs/issues/803

dougiesquire:
Your link shows it as a reduction method (which makes sense) but it your table you also have it listed in the "frequency" row

aidanheerdegen:
Ok, now is the time to fix this, before we have another 7 years of inconsistent outputs.

Would this require concomitant changes in the [intake catalogue AccessEsm15Builder](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/5a583c9ebccaa712e114960316ee40475fe4a7e0/src/access_nri_intake/source/builders.py#L649) @charles-turner-1?

aidanheerdegen:
I decided to roll with it as-is, and run the `addmeta` tool for each group of frequencies in turn, which makes for a much more complex post-processing task, but it can be collapsed at a later date if the filenames are modified to be compliant with our schema.

PR #526, opened 2026-03-03
Change wombatlite ocean diagnostic file naming convention

aidanheerdegen:
**1. Summary**:

Change one of the ocean diagnostic output filename profiles to match the [frequency schema](https://github.com/ACCESS-NRI/schema/blob/main/au.org.access-nri/model/output/file-metadata/2-0-0/global/frequency.json)

**2. Issues Addressed:**

- #224 

- [ ] input workflow:

**4. Ad-hoc Testing**

**5. CI Testing**

**6. Reproducibility**

- [x] Yes

**7. Performance**

- [ ] Yes
- [x] No

**8. Manifests**

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

```yaml
manifest:
  reproduce:
    exe: false
    input: false
    restart: false
runlog:
  enable: false
```

**9. Documentation**

Is the documentation updated?
- [ ] Yes
- [ ] N/A

**10. Merge Strategy**

- [ ] Merge commit
- [ ] Rebase and merge
- [ ] Squash

### DIFF
--- a/ocean/diagnostic_profiles/diag_table_detailed
+++ b/ocean/diagnostic_profiles/diag_table_detailed
@@ -10,620 +10,620 @@ ACCESS-ESM_1p6
 
 # static 2d grid data
 
-"ocean-2d-area_t", -1, "months", 1, "days", "time"
-"ocean_model", "area_t", "area_t", "ocean-2d-area_t", "all", "none", "none", 2
+"ocean-2d-area_t-fx", -1, "months", 1, "days", "time"
+"ocean_model", "area_t", "area_t", "ocean-2d-area_t-fx", "all", "none", "none", 2
 
-"ocean-2d-area_u", -1, "months", 1, "days", "time"
-"ocean_model", "area_u", "area_u", "ocean-2d-area_u", "all", "none", "none", 2
+"ocean-2d-area_u-fx", -1, "months", 1, "days", "time"
+"ocean_model", "area_u", "area_u", "ocean-2d-area_u-fx", "all", "none", "none", 2
 
-"ocean-2d-geolat_c", -1, "months", 1, "days", "time"
-"ocean_model", "geolat_c", "geolat_c", "ocean-2d-geolat_c", "all", "none", "none", 2
+"ocean-2d-geolat_c-fx", -1, "months", 1, "days", "time"
+"ocean_model", "geolat_c", "geolat_c", "ocean-2d-geolat_c-fx", "all", "none", "none", 2
 
-"ocean-2d-geolat_t", -1, "months", 1, "days", "time"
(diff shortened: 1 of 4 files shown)


dougiesquire (approved):
LGTM2

ocean/diagnostic_profiles/source_yaml_files/diag_table_wombatlite_source.yaml
```diff
@@ -77,11 +77,11 @@ global_defaults:
...
-        years: yearly
-        months: monthly
-        days: daily
-        hours: hourly
-        none: snap  # careful! will apply to both reduction_method and regional_section
+        years: yr
+        months: mon
+        days: day
+        hours: hr
+        none: fx  # careful! will apply to both reduction_method and regional_section
```
aidanheerdegen: Is this correct @dougiesquire ? What about the warning about `reduction_method` and `regional_section`?

ocean/diagnostic_profiles/source_yaml_files/diag_table_wombatlite_source.yaml
```diff
@@ -77,11 +77,11 @@ global_defaults:
...
         'False': snap
         average: mean
         'True': mean
```
aidanheerdegen: I haven't done anything with these because I don't know how they're used.

aidanheerdegen:
> This PR should probably keep the files consistent (ie update the substitutions in the other files also) and we can discuss their fate in a separate issue?

Probably. I didn't change the others because I wanted feedback if what I had done was correct.

Specifically https://github.com/ACCESS-NRI/access-esm1.6-configs/pull/526#discussion_r2887399823 and https://github.com/ACCESS-NRI/access-esm1.6-configs/pull/526#discussion_r2887401878

ocean/diagnostic_profiles/source_yaml_files/diag_table_wombatlite_source.yaml
dougiesquire: I'm not sure I'd worry about the warning in the comment because we don't use regional sections.

Regarding `none: fx`:
`reduction_method`=`none` means a snapshot, not fixed frequency. But we don't have any outputs that use `reduction_method`=`none` and include `reduction_method` in the filename. So this does nothing. Still, possibly better to remove line 84 altogether?

dougiesquire: > Still, possibly better to remove line 84 altogether?

Sorry, this is bad advice. I think leave as `none: snap`. Then, for example, if we were to add daily snapshots, they would end up as something like:

```
ocean-scalar-1day-snap-ym_2026_03.nc
```

PR #523, opened 2026-03-02
Test addmeta

aidanheerdegen:
**1. Summary**:

Added postscript changes to run addmeta tool for global metadata to make data conform to [ACCESS Output Data Spec](https://access-output-data-specifications.readthedocs.io/en/latest/).

**2. Issues Addressed:**

-

**3. Dependencies (e.g. on payu, or model)**

This change requires changes to (note pull request(s) where relevant):
- [ ] workflow manager (payu):
- [ ] model deployment (ACCESS-ESM1.6):
- [ ] model component or library dependency:
- [ ] input workflow:
- [x] post-processing workflow:

**4. Ad-hoc Testing**

No changes were made to model configuration, so no testing was done to ensure bit repro.

**5. CI Testing**

- [ ] `!test repro` has been run

No.

**6. Reproducibility**

Is this reproducible with the previous commit? (If not, why not?)
- [X] Yes
- [ ] No - `!test repro commit` has been run. 

**7. Performance**

Has the model performance (say, throughput of model-years/wall-day) changed? 
- [ ] Yes
- [X] No
- [ ] N/A (if selected, please add a brief explanation why performance testing is not necessary for this PR)

If yes, provide the numbers from your testing. Is the performance better or worse?

**8. Manifests**

Have you changed the executable, the input files and/or the restart files?
- [ ] Yes
- [X] No

If yes, have you updated the manifests?
- [ ] Yes
- [X] No

To update the manifests, run payu setup (in a cloned copy of your feature branch) with reproducibility tests turned off:

```yaml
manifest:
  reproduce:
    exe: false
    input: false
    restart: false
runlog:
  enable: false
```
Then commit the newly created manifest files (under manifests/) only to the branch for this PR.

**9. Documentation**

Is the documentation updated?
- [ ] Yes
- [ ] N/A

Not updated yet.

**10. Merge Strategy**

- [ ] Merge commit
- [ ] Rebase and merge
- [X] Squash

PR #603, opened 2026-03-18
Added config files for time & time_bnds/bounds.

joshuatorrance:
### Summary

This PR merges with the branch for the test-addmeta PR #523. So I've glossed over most of the details in the PR template.

### What has changed?

Added two config files to add attrs for time and remove attrs for time_bnds/time_bounds.

Unresolved issues:
- The `units` attribute of `time` has not been set here as reference time is not guaranteed to be `0000-01-01 00:00:00`
  - This leaves a mild inconsistency between realms as some have `hh:mm:ss` in the reference time, others have just `hh:mm`
- Secondary *_bnds dimension names are changed from `d2` & `nv` to `bnds`
  - nothing else uses the names `d2` or `nv` in the datasets I've looked through
  - ocean files have `nv` as a variable too, this is also renamed

### Why was this done?

To allow addmeta to unify the attrs present for time/time_bnds for all realms.


joshuatorrance:
> I'm unsure whether `units` should be included here or not.
> 
> In theory units could have a unknown reference time, `days since 1234-01-01 10:31:59`
> 
> If we're 100% certain than all time units coming out of ESM1.6 will always have `days since 0001-01-01 00:00:00` then we could set that here as well.

From talking with some folks it sounds like we don't always use the same reference datetime. Thus we cannot easily set the units with `addmeta`. Maybe some Jinja sorcery can be used to extract the datetime from the `units` string and format it similarly?

joshuatorrance:
> Jinja sorcery

```python
s = "{{ t.replace(' 00:00:00', '').replace(' 00:00', '') if ' 00:00:00' == t[-9:] or ' 00:00' == t[-6:] in t else t }}"
d = {"t": "days since 0001-02-03 00:00"}

jinja2.Template(s).render(d)
>>> 'days since 0001-02-03'
```

This would work *if* we were able to refer to file attrs as templates variables in `addmeta`.

While possible I'm inclined to just let the units be mildly inconsistent - e.g.
- atmosphere uses `days since 0001-01-01 00:00` and
- ice and ocean use `days since 0001-01-01 00:00:00`

addmeta/time.yaml
aidanheerdegen, line 1: I am fine with this as a separate file, but just to note with the changes to being able to use a single file, this could be include in the `dataspec.yaml` file as well.

aidanheerdegen (approved):
LGTM, with a single comment/question.

Issue #642, opened 2026-03-30
Archive +CN configs


anton-seaice:
Following on from the stand-up meething this morning, for ease of maintenance, we can archive the CN only configurations

I suggest we create a new prefix (e.g. `archive-`) and rename the [existing](https://github.com/ACCESS-NRI/access-esm1.6-configs/branches/all?query=%2BCN) `dev-` configs to have an `archive-` prefix. Ideally if there is a word for `archive-` which is at the end of the alphabet that would be great, so they are at the bottom of the list of branches in this repository.

We should also create a record of where the data is from existing "CN only" experiments - possibly just in this Issue ?

This reduces the number of supported configurations by ~9, which makes maintenance simpler / less error prone.

@aidanheerdegen @ccarouge @blimlim

anton-seaice:
> rename the [existing](https://github.com/ACCESS-NRI/access-esm1.6-configs/branches/all?query=%2BCN) `dev-` configs to have an `archive-` prefix.

possibly archiving them to a tag with the `archive-` prefix is neater? tags are a bit more hidden away

blimlim:
> possibly archiving them to a tag with the archive- prefix is neater?

This idea sounds good!

anton-seaice:
@aidanheerdegen - there's about ~9 configs that are no longer the focus for development, but some spinpup runs were done with them. I'm suggesting creating tags with the `archive-` prefix for them and deleting the branches, how does that sound ?

aidanheerdegen:
> I'm suggesting creating tags with the `archive-` prefix for them and deleting the branches, how does that sound ?

Sounds fine to me. I'd suggest adding a note in the README to this effect so people know how to find them if they need to.

anton-seaice:
Ok archived these:

  origin/dev-1pctCO2+CN
  origin/dev-1pctCO2-bgc+CN
  origin/dev-1pctCO2-rad+CN
  origin/dev-4xCO2+concentrations+CN
  origin/dev-amip+CN
  origin/dev-flat10+CN
  origin/dev-historical+concentrations+CN
  origin/dev-preindustrial+concentrations+CN
  origin/dev-preindustrial+emissions+CN

ccarouge:
[...]

PR #691, opened 2026-04-14
Docs: updates for new and archived branches

anton-seaice:
Docs update for #642 - this updates the branches listed and notes about +CN configurations which are archived.

### DIFF
--- a/documentation/docs/pages/index.md
+++ b/documentation/docs/pages/index.md
@@ -30,14 +30,9 @@ Each configuration in [github.com/ACCESS-NRI/access-esm1.6-configs](https://gith
 
 where release signifies this is the release branch that is tested, versioned and ready for use, scenario is the base experimental design with optional modifiers. All configurations are global with nominal 1 degree resolution.
 
-Some examples of possible values of the specifiers:
+`scenario` is based around the CMIP experiment identifiers, examples include `historical`, `preindustrial`, `4xCO2`.
 
-- scenario: 
-- modifier: `concentrations`, `emissions`, `CN`
-
-Scenario is based around the CMIP experiment identifiers, examples include `historical`, `preindustrial`, `4xCO2`.
-
-The modifier can be `concentrations` or `emissions` to describe the CO2 cycling protocol, or `CN` to include the carbon and nitrogen cycles but excluding the phosphorus cycle in the land.
+The `modifier` can be `concentrations` or `emissions` to describe the CO2 cycling protocol, or `CN` to include the carbon and nitrogen cycles but excluding the phosphorus cycle in the land.
 
 These names do not exactly match [CMIP7 experiment names](https://airtable.com/embed/apphXCUgASIeT6jCz/shrCs1cSWzQRV0v4i/tblbT6XAdQYOCMXu7), we use `+concentrations` modifier for concentration driven experiments (these are more similar to CMIP6 experiments, and do not have `esm` in their names) and `+emissions` for emissions driven experiments (those with the `esm-` prefix in their CMIP7 name).
 
--- a/documentation/docs/pages/index.md
+++ b/documentation/docs/pages/index.md
@@ -32,10 +32,12 @@ where release signifies this is the release branch that is tested, versioned and
 
 Some examples of possible values of the specifiers:
 
-- scenario: `historical`, `preindustrial`, `4xCO2`
-- modifier: `concentrations`, `emissions`
+- scenario: 
+- modifier: `concentrations`, `emissions`, `CN`
 
-where scenario is typically a CMIP experiment identifier and `concentrations` and `emissions` describe the CO2 cycling protocol.
+Scenario is based around the CMIP experiment identifiers, examples include `historical`, `preindustrial`, `4xCO2`.
+
+The modifier can be `concentrations` or `emissions` to describe the CO2 cycling protocol, or `CN` to include the carbon and nitrogen cycles but excluding the phosphorus cycle in the land.
 
 These names do not exactly match [CMIP7 experiment names](https://airtable.com/embed/apphXCUgASIeT6jCz/shrCs1cSWzQRV0v4i/tblbT6XAdQYOCMXu7), we use `+concentrations` modifier for concentration driven experiments (these are more similar to CMIP6 experiments, and do not have `esm` in their names) and `+emissions` for emissions driven experiments (those with the `esm-` prefix in their CMIP7 name).
 
@@ -84,10 +86,8 @@ These were `+CN` only, i.e. including the carbon and nitrogen cycles but excludi
 - [archive-preindustrial+concentrations+CN](https://github.com/ACCESS-NRI/access-esm1.6-configs/releases/tag/archive-preindustrial%2Bconcentrations%2BCN)
 - [archive-preindustrial+emissions+CN](https://github.com/ACCESS-NRI/access-esm1.6-configs/releases/tag/archive-preindustrial%2Bemissions%2BCN)
 
-If you would like to run these archived configurations, please use the 
-[ACCESS-Hive forum for support](https://forum.access-hive.org.au/new-topic?category=esm&tags=access-nri-help) on how to 
-update these configurations up with the latest changes in the configurations under active development.
-
+If you need assistance to use git to update these configurations, please ask on the 
+[ACCESS-Hive forum](https://forum.access-hive.org.au/new-topic?category=esm&tags=access-nri-help).
 
 #### How to use this repository to run a model
 
--- a/documentation/docs/pages/index.md
+++ b/documentation/docs/pages/index.md
@@ -32,22 +32,27 @@ where release signifies this is the release branch that is tested, versioned and
 
 Some examples of possible values of the specifiers:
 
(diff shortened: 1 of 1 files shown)


documentation/docs/pages/index.md
```diff
@@ -28,26 +28,34 @@ Each configuration in [github.com/ACCESS-NRI/access-esm1.6-configs](https://gith
...
 Some examples of possible values of the specifiers:
 
-- scenario: `historical`, `preindustrial`, `ssp126`
-- modifier: `concentration`, `interactiveC`, `noLUC`
+- scenario: `historical`, `preindustrial`, `4xCO2`
+- modifier: `concentration`, `emissions`
+
+where scenario is typically a CMIP experiment identifier and concentration and emissions describe the CO2 cycling protocol.
+
+These names do not exactly match [CMIP7 experiment names](https://airtable.com/embed/apphXCUgASIeT6jCz/shrCs1cSWzQRV0v4i/tblbT6XAdQYOCMXu7), we use `+concentration` modifier for concentration driven experiments (these are more similar to CMIP6 experiments, and do not have `esm` in their names) and `+emissions` for emissions driven experiments (those with the `esm-` prefix in their CMIP7 name).
```
anton-seaice: We can revisit after #697

documentation/docs/pages/index.md
```diff
@@ -28,26 +28,34 @@ Each configuration in [github.com/ACCESS-NRI/access-esm1.6-configs](https://gith
...
-- [dev-preindustrial+concentrations](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-preindustrial%2Bconcentrations)
-- [dev-preindustrial+emissions](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-preindustrial%2Bemissions)
-- [dev-4xCO2+concentrations](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-4xCO2%2Bconcentrations)
+
 - [dev-1pctCO2](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-1pctCO2)
+- [dev-1pctCO2-bgc](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-1pctCO2-bgc)
+- [dev-1pctCO2-rad](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-1pctCO2-rad)
+- [dev-4xCO2+concentrations](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-4xCO2%2Bconcentrations)
 - [dev-amip](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-amip)
+- [dev-flat10](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-flat10)
```
ccarouge, line 50: Should we rename to `dev-flat10+emissions` considering #697 and do the renaming before pushing this documentation change so we can change in the doc at the same time?

documentation/docs/pages/index.md
```diff
@@ -63,6 +71,22 @@ Where changes are made, they should be applied to all these branches (where rele
 ](https://github.com/ACCESS-NRI/access-esm1.6-configs/compare/dev-preindustrial+concentrations..dev-1pctCO2
 )
 
+There were some configurations previously under development that have now been archived. 
+These were `+CN` only, without a phosporus cycle in the land and are available for reference:
```
ccarouge:
```suggestion
These were `+CN` only, i.e. including the carbon and nitrogen cycles but excluding the phosphorus cycle in the land and are available for reference:
```
Or do we want a table of all modifiers somewhere with their definition? Might be handy?

documentation/docs/pages/index.md
```diff
@@ -63,6 +71,22 @@ Where changes are made, they should be applied to all these branches (where rele
...
+- [archive-1pctCO2+CN](https://github.com/ACCESS-NRI/access-esm1.6-configs/releases/tag/archive-1pctCO2%2BCN)
+- [archive-4xCO2+concentrations+CN](https://github.com/ACCESS-NRI/access-esm1.6-configs/releases/tag/archive-4xCO2%2Bconcentrations%2BCN)
+- [archive-flat10+CN](https://github.com/ACCESS-NRI/access-esm1.6-configs/releases/tag/archive-flat10%2BCN)
+- [archive-historical+concentrations+CN](https://github.com/ACCESS-NRI/access-esm1.6-configs/releases/tag/archive-historical%2Bconcentrations%2BCN)
+- [archive-preindustrial+concentrations+CN](https://github.com/ACCESS-NRI/access-esm1.6-configs/releases/tag/archive-preindustrial%2Bconcentrations%2BCN)
+- [archive-preindustrial+emissions+CN](https://github.com/ACCESS-NRI/access-esm1.6-configs/releases/tag/archive-preindustrial%2Bemissions%2BCN)
+
+If you would like to run these archived configurations, please use the 
+[ACCESS-Hive forum for support](https://forum.access-hive.org.au/new-topic?category=esm&tags=access-nri-help) on how to 
+update these configurations up with the latest changes in the configurations under active development.
```
ccarouge: Do we really want to provide that kind of support for archived configurations?

ccarouge (changes requested):
A few comments. While writing I think we should provide a table of all the modifiers we are using and what they mean (instead of examples and in text explanations). Could do the same for the scenario, with the config name and the equivalent CMIP7 one when it exists.

documentation/docs/pages/index.md
ccarouge: Would we be better providing generic instructions and let the community figure it out? Instead of saying people will get personalised help whenever they want to revive an archive configuration.

documentation/docs/pages/index.md
```diff
@@ -28,26 +28,34 @@ Each configuration in [github.com/ACCESS-NRI/access-esm1.6-configs](https://gith
...
 
-where release signifies this is the release branch that is tested, versioned and ready for use, scenario is the base experimental design with optional modifiers. All configurations are assumed to be global extent with nominal 1 degree resolution.
+where release signifies this is the release branch that is tested, versioned and ready for use, scenario is the base experimental design with optional modifiers. All configurations are global with nominal 1 degree resolution.
 
 Some examples of possible values of the specifiers:
 
-- scenario: `historical`, `preindustrial`, `ssp126`
-- modifier: `concentration`, `interactiveC`, `noLUC`
+- scenario: `historical`, `preindustrial`, `4xCO2`
+- modifier: `concentration`, `emissions`
```
blimlim:
```suggestion
- modifier: `concentrations`, `emissions`
```

documentation/docs/pages/index.md
```diff
@@ -28,26 +28,34 @@ Each configuration in [github.com/ACCESS-NRI/access-esm1.6-configs](https://gith
...
+
+where scenario is typically a CMIP experiment identifier and concentration and emissions describe the CO2 cycling protocol.
```
blimlim:
```suggestion
where scenario is typically a CMIP experiment identifier and `concentrations` and `emissions` describe the CO2 cycling protocol.
```

documentation/docs/pages/index.md
blimlim:
```suggestion
These names do not exactly match [CMIP7 experiment names](https://airtable.com/embed/apphXCUgASIeT6jCz/shrCs1cSWzQRV0v4i/tblbT6XAdQYOCMXu7), we use `+concentrations` modifier for concentration driven experiments (these are more similar to CMIP6 experiments, and do not have `esm` in their names) and `+emissions` for emissions driven experiments (those with the `esm-` prefix in their CMIP7 name).
```

documentation/docs/pages/index.md
anton-seaice: I think fix this after #697

documentation/docs/pages/index.md
anton-seaice: I didn't think anyone would take up this offer, so wasn't worried too much.

I could generalise a bit, ...

_If you need assistance in using git to update these configurations, please use the ACCESS-Hive forum_ ?

ccarouge: Yes, I think that's better. Maybe "ask on the ACCESS-Hive forum" instead of "use".

ccarouge:
@anton-seaice did you miss this: "I think we should provide a table of all the modifiers we are using and what they mean (instead of examples and in text explanations). " Sorry, left it as an overall comment. What do you think?

anton-seaice:
> @anton-seaice did you miss this: "I think we should provide a table of all the modifiers we are using and what they mean (instead of examples and in text explanations). " Sorry, left it as an overall comment. What do you think?

Possibly do this after #697 ?

But then , the only current ones are +emissions and +concentrations, so it's sort of simple.

ccarouge:
> But then , the only current ones are +emissions and +concentrations, so it's sort of simple.

There is also `+CN`. I would list the archived ones, because we know they exist and prevent us using a different one if we need to do the same thing again.

Happy to do after #697

anton-seaice:
Ok refined the wordings a little @ccarouge

ccarouge (approved):
LGTM

Issue #698, opened 2026-04-15
Change preindustrial+concentrations restart


blimlim:
For the rerun of the preindustrial+concentrations configuration, we would like to swap the restart to a copy of `/g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/Ndep2-PI-CNP-concentrations/restart170` with the year changed to 1.

This restart comes from @JhanSrbinovsky's test run with the Ndep v2 file, and was started from `restart099` of @MartinDix's test run using the Ndep v2 file.

PR #700, opened 2026-04-15
preindustrial+concentrations restart: don't change year

blimlim:
**1. Summary**:
Following further discussion with @JhanSrbinovsky and @rml599gh, we would like to use the same restart from #699 but without the change to the restart year.

This PR updates the restart to use a new copy, without the date change.

**2. Issues Addressed:**

- #698 

- [ ] input workflow:

**4. Ad-hoc Testing**

**5. CI Testing**

**6. Reproducibility**

- [x] Yes

**7. Performance**

- [ ] Yes
- [ ] No
- [x] N/A (if selected, please add a brief explanation why performance testing is not necessary for this PR)

**8. Manifests**

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

```yaml
manifest:
  reproduce:
    exe: false
    input: false
    restart: false
runlog:
  enable: false
```

**9. Documentation**

Is the documentation updated?
- [ ] Yes
- [x] N/A

**10. Merge Strategy**

- [ ] Merge commit
- [ ] Rebase and merge
- [x] Squash


blimlim:
!test repro

blimlim:
> ❌ The Bitwise Reproducibility Check Failed ❌

Not expected. I'll try to see what happened

JhanSrbinovsky (commented):
Thanks Spencer

JhanSrbinovsky:
what happened this was just the start date right? not even that, just a year number.

blimlim:
Yeah, only the restart year has changed which is not expected to change results. The checksums are identical for all fields other than `age_global`:

```
diff -c 2 checksum1 checksum2

      ],
      "age_global": [
!       "-948087343509176042"
      ],
      "frazil": [
--- 218,222 ----
      ],
      "age_global": [
!       "-8010627089325603220"
      ],
      "frazil": [

```

blimlim:
The following shows the column mean ocean age tracer for 1 month runs starting from each restart:

Non year 1 start:
download-91

Year 1 start:
download-92

MOM appears to completely reset the age tracer when starting from year 1. This might be something we don't want to ever happen, I'll make a separate issue for discussion.

blimlim:
!test repro commit

rml599gh:
I'm thinking that whether or not we change the date for this configuration depends on what we want it to represent if we are keeping the picontrol that does have the restart date change in.

blimlim:
@JhanSrbinovsky, @rml599gh, this PR swaps the restart for the preindustrial+concentrations run to a copy of `/g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/Ndep2-PI-CNP-concentrations/restart170` without the date change. If you are happy to review this that would be great!

I'll add the MOM namelist change to a separate PR which will let the change be copied to all the configurations

JhanSrbinovsky:
phew, I'm glad my memory wasn't playing tricks on me: That was the restart we used before from:

```
/g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2026.04.15/README 
```

Can you clarify, at the start of the ocean age tracer discussion - there was a comment that I read as "even though the age tracer changes, reproducibility shattered, all other physical characteristics are the same." Was that not the case?

rml599gh (approved):
OK - I think I've got this straight in my mind now. This takes the end of Jhan's picontrol run and puts it into the dev config instead of the test one - which is what we want.

blimlim:
Thanks @rml599gh and @JhanSrbinovsky 

>"even though the age tracer changes, reproducibility shattered, all other physical characteristics are the same." Was that not the case?

Yeah that's correct – the ocean age tracer variable changed, but all other aspects of the physical state in the ocean, atmosphere and sea ice weren't affected. 


>  This takes the end of Jhan's picontrol run and puts it into the dev config instead of the test one - which is what we want.

That's right, It's taking the restart from the end of Jhan's run and putting it into the dev config

Issue #703, opened 2026-04-15
Ocean age tracer gets reset when starting from year 1


blimlim:
PR #473 attempted to resolve the reproducibility issues caused by swapping the restart year to year 1. However the failing reproducibility check in #700 suggests that changing the date to 1/1/1  has an additional impact on the ocean age tracer. MOM appears to reset the ocean age when starting from 1/1/1. The following show column mean ocean age for two runs using the same restart, but where one has the date reset to 1/1/1:

Starting from year 272:
<img width="574" height="453" alt="Image" src="https://github.com/user-attachments/assets/471884ad-0eb6-45cc-9098-08370f210087" />

Starting from same restart, with date reset to 1/1/1:
<img width="579" height="453" alt="Image" src="https://github.com/user-attachments/assets/64316aab-9706-4fb8-b622-01ba72d93bbc" />

While this doesn't affect the rest of the model physical state, we lose any spinup of the ocean age.

The following shows the start dates for the preindustrial restarts in `/g/data/vk83` (excluding the new ones from the last couple of days). There's been one point where we reset the start date to 1/1/1 during the spinup/control runs:
end_date: 0101-01-01 00:00:00
end_date: 1092-01-01 00:00:00
end_date: 1843-01-01 00:00:00
end_date: 1843-01-01 00:00:00
end_date: 2168-01-01 00:00:00
end_date: 2198-01-01 00:00:00
end_date: 2198-01-01 00:00:00
end_date: 0602-01-01 00:00:00
end_date: 1092-01-01 00:00:00
end_date: 0402-01-01 00:00:00
end_date: 2198-01-01 00:00:00
end_date: 2108-01-01 00:00:00
end_date: 2145-01-01 00:00:00
end_date: 2146-01-01 00:00:00
end_date: 2676-01-01 00:00:00
**end_date: 0001-01-01 00:00:00**
end_date: 0101-01-01 00:00:00

The reset was applied to `/g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2026.02.20`, which formed the starting point of Tilo's emissions driven picontrol (`/scratch/p66/txz599/access-esm/archive/pi-esm-01-expt-2c021aa7/`) and Martin's test of the Ndep v2 dataset. From the first month of output, both of these runs appear to have the age tracer reset:

[...]

PR #707, opened 2026-04-21
Increase age tracer limit

blimlim:
**1. Summary**:
In #703 we found that changing the restart date could result in the age tracer being overwritten. With the current settings, [MOM5 limits the age tracer to be less than the model time](https://github.com/ACCESS-NRI/MOM5/blob/01f92d10dff315b369b8b668971944d8ca71c22e/src/mom5/ocean_tracers/ocean_tracer.F90#L2509-L2511). As a result, changing the restart date to 1/1/1 for example caused the age tracer to be reset to 0 everywhere.

It is possible to completely deactivate the age limiting in the namelist, by setting `limit_age_tracer = .false.`. However this also deactivates limiting of negative values, resulting in [negative ages appearing in the output](https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/703#issuecomment-4257965416).

An alternative is to leave the age limiting active, and set a very high maximum age limit `age_tracer_max_init` that hopefully would never be reached. This PR sets the maximum age limit to 1000000 years, which I hope is large enough to not be reached in a model run.

**2. Issues Addressed:**

- #703

- [ ] input workflow:

**4. Ad-hoc Testing**

The following plot of the zonal maximum age tracer is from a run using `/g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/pre-industrial/restart/2026.02.20` which had its restart date reset to 1/1/1

There appears to be no limiting occuring:
download-94

**5. CI Testing**

**6. Reproducibility**

- [ ] Yes
- [x] No - `!test repro commit` has been run. 

**7. Performance**

- [ ] Yes
- [ ] No
- [x] N/A (if selected, please add a brief explanation why performance testing is not necessary for this PR)

This option should not affect performance

**8. Manifests**

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

```yaml
manifest:
  reproduce:
    exe: false
    input: false
    restart: false
runlog:
  enable: false
```

**9. Documentation**

Is the documentation updated?
- [ ] Yes
- [ ] N/A

**10. Merge Strategy**

- [x] Merge commit
- [ ] Rebase and merge
- [ ] Squash


blimlim:
!test repro

dougiesquire (dismissed):
Great work tracking this down @blimlim. LGTM

blimlim:
!test repro commit

dougiesquire (approved):

blimlim:
Thanks @dougiesquire for the review!

Just to note, the only checksum change is for the age tracer. I wasn't expecting this as the current restart file hasn't had it's date modified. However, it appears that the age tracer limiting still has an effect on the age tracer. The following show month mean, and first timestep differences for runs with and without this change:

download-96
download-95

dougiesquire:
That's odd. So it was previously being limited even when the date wasn't modified?

blimlim:
> So it was previously being limited even when the date wasn't modified?

It's looking like it. The differences in the first time step are all one sign, and so it seems like the limiting was still somehow having an effect:

download-97

blimlim:
In the discussion in #703, we saw that the numerics can lead to negative ocean age, if a lower limit is not applied. I suspect that the numerics may also cause increases in age which are greater than the model timestep. These errors can be much larger than noise (e.g. in the above plots, the limiting causes a difference of 1.3h after a 1h timestep).

However, removing the upper limit appears to be the only way to prevent problems when changing the date, and so we've decided that we will merge this change.

blimlim:
!cherry-pick d284c861b13e596ae4559c33b1fe053470920cb5 into dev-4xCO2+concentrations dev-1pctCO2 dev-1pctCO2-bgc dev-1pctCO2-rad dev-flat10 dev-preindustrial+emissions dev-historical+emissions dev-historical+concentrations

Issue #601, opened 2026-03-16
Running an ensemble


MartinDix:
Want to be able to run an emissions driven historical ensemble starting from every 20 years of the control run.

Requires modifying the date of the restart files and initialising the wood product fields in the UM restart.

blimlim:
Following further discussion with @aidanheerdegen and @anton-seaice, we'd like to change the approach a little. A concern with updating the restarts on the fly, is that the md5 hashes in the manifests are not accurate to what the model used. 

The proposed updated workflow would be:

* User clones experiment/configuration
* User runs a script which:
   * Makes a copy of a selected restart file somewhere (maybe in the archive directory?)
   * Modifies the date in the copy and sets the initial wood product
   * Updates the config.yaml to point to the modified restart
   * Adds a commit recording the paths used, modifications made, any relevant environment versions

* User runs experiment

We might want to consider setting the 1850 wood thinning data as part of the above script, and changing the update_thinning script to run as an archive script along with `update_landuse_driver.sh` – this would also be required for the md5 hashes to match the restarts actually used in each run.

anton-seaice:
Hi @rml599gh 

could you confirm how many of each experiment wanted for each ensemble? and which restart years to start from please ?

e.g. every 20th year with the first being `/scratch/p66/jxs599/access-esm/archive/Ndep2-PI-CNP-emissions-dev-preindustrial+emissions-14-04-26-cfcbc3a9/restart098/`  ?

My reading of this version confusing link is that names are unchanged from cmip6, so for us should be r1i1p1f1 through r30i1p1f1

https://wcrp-cmip.github.io/cmip7-guidance/docs/CMIP7/Global_Attributes/#table-2-required-global-attributes

rml599gh:
Hi @anton-seaice , @blimlim 
The aim is to do at least 30 members for the emissions-driven historical, starting from every 20 years. The first member (the one I've already run) started from /scratch/p66/jxs599/access-esm/archive/Ndep2-PI-CNP-emissions-dev-preindustrial+emissions-14-04-26-cfcbc3a9/restart199 so the next would be restart219, restart239 .... These restarts are 100 years later than you might expect because we are taking the first 100 years as spin-up and will only publish the esm-picontrol from output100 (model year 101).
Although we want to run 30 eventually, we assume it is only realistic to have around 5 running concurrently. Also, some will be run under different project codes.
[...]

PR #741, opened 2026-05-11
Add ensemble generation scripts to historical configurations

blimlim:
**1. Summary**:

This PR adds scripts to setup the initial restart file for historical experiments, with the aim of making it easy to create new ensemble members from different preindustrial restarts.

The scripts perform the following steps:
1. Set the start year to 1850 in the restart files, based on the steps in https://access-nri.github.io/access-esm1.6-configs/inputs/restarts/#changing-the-date-of-restart-files. This step also updates the UM namelist files with the correct date. This is required as the scripts are run after the payu setup stage, which normally would set the correct values in the namelists.
2. `insert_external_woodprod_CNP.py` script to set the initial 20PgC wood product.
3. After the run is complete, the modified CICE restart is manually removed. This is required as payu only deletes files based on those found in the directory specified under the `restart` field in the config.yaml. 

Modifications to the restart files in the above steps are made in a temporary directory rather than in place, to prevent accidental modification of the files in vk83.

**2. Issues Addressed:**

- #601 

- [ ] input workflow:

**4. Ad-hoc Testing**

**5. CI Testing**

**6. Reproducibility**

- [ ] Yes

**7. Performance**

- [ ] Yes
- [ ] No

These scripts only modify the restart for the initial run and aren't expected to affect performance.

**8. Manifests**

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

```yaml
manifest:
  reproduce:
    exe: false
    input: false
    restart: false
runlog:
  enable: false
```

**9. Documentation**

Is the documentation updated?
- [ ] Yes
- [ ] N/A

**10. Merge Strategy**

- [ ] Merge commit
- [ ] Rebase and merge
- [ ] Squash


blimlim:
!test repro commit

scripts/set_restart_1850.sh
```diff
@@ -0,0 +1,59 @@
...
+set -eu
+
+# This is only going to work if the payu run numbers hasn't been changed
+if [[ $PAYU_CURRENT_RUN != 0 ]]; then
+  echo "Skipping restart changes"
+  exit 0
+fi
+
+cp work/atmosphere/restart_dump.astart $TMPDIR
+python ~access/apps/pythonlib/umfile_utils/access_cm2/change_dump_date.py $TMPDIR/restart_dump.astart <<< "1850 01 01"
```
anton-seaice: @MartinDix suggested copying this into the scripts folder, which might work


but where are these available from in the change_dump_date script ?

```
import umfile
from um_fileheaders import *
```

blimlim: Good point, those are coming from `/g/data/access/projects/access/apps/pythonlib/umfile_utils/access_cm2/umfile.py` and `/g/data/access/projects/access/apps/pythonlib/umfile_utils/access_cm2/um_fileheaders.py`.

I'll try out copying in the refactored version which doesn't have those dependenceis: https://github.com/ACCESS-NRI/umfile_utils/blob/main/src/umfile_utils/change_dump_date.py

anton-seaice: Should we install umfile_utils into the payu container, or `vk83` ?

blimlim: Would we need to possibly do both, or does that get too messy? It's likely to be installed in an environment containing several tools for handling the UM files: https://github.com/ACCESS-NRI/containerised-environments-infra/issues/38

But this apparently wouldn't be useable from a payu userscript, something to do with trying to launch another container from the payu one.

anton-seaice: Sounds like we should resurrect the [calendar:start](https://github.com/ACCESS-NRI/access-esm1.6-configs/commit/9fa7b7f4545bc6764bb175dcdf4b1b835f0fe8cf)  config lines and implement them in payu !

blimlim:
Tested a 1 year run with the config in this PR, with restart199 from the esm-piControl experiment (`/scratch/p66/jxs599/access-esm/archive/Ndep2-PI-CNP-emissions-dev-preindustrial+emissions-14-04-26-cfcbc3a9/restart199`).

We get identical answers as @rml599gh's run, which used the same restart but applied the date changes and initial wood product manually: 
```
> nccmp -dgf aiihca.pa-185012_mon.nc /g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/esm-historical-01/output000/atmosphere/netCDF/aiihca.pa-185012_mon.nc
>
```

blimlim:
@NortonAlex, this PR adds in a copy of the initial wood product script from ` /g/data/p66/ajn563/ACCESS-ESM/ESM1.6/configs/insert_external_woodprod_CNP.py`. There are some small changes for passing it the filepaths from the configuration. If you are happy to review those changes that would be great!

scripts/initial_wood_product.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -eu
+
+# Set the size of the initial wood carbon nitrogen and phosphorous pools
+# for an emissions driven historical simulation. This modifies the atmospheric
```
NortonAlex: This same script will be used for concentration-driven and emissions-driven, correct? Just want to make this comment clear.

blimlim: Thanks @NortonAlex, good point. I've updated this part of the comment to `Set the size of the initial wood carbon nitrogen and phosphorous pools for historical simulations.`

blimlim:
!test repro commit

blimlim:
!test repro commit

scripts/initial_wood_product.sh
NortonAlex: Super

NortonAlex:
@blimlim any idea why the checks are failing on this?

blimlim:
>@blimlim any idea why the checks are failing on this?
It looks like there's an issue with running the date change scripts when the original restart already has year 1850.

I think in this case, both payu and the `restart_1850_cleanup.sh` script attempt to delete the 1850 restart causing a file not found error.

In the normal case when the original restart has a different year, payu fails to clean up the modified 1850 restart, which is why the `restart_1850_cleanup.sh` was added.

I'll try and get in a workaround.

blimlim:
!test repro

blimlim:
Following discussion with @aidanheerdegen and @anton-seaice, a change in approach was decided on. In the initial implementation, the date change and initial wood product scripts were run automatically after the `payu setup` step. An issue was that payu calculates the restart manifests during the setup stage, and so the md5 hashes in the manifests did not match the files that the model actually started from.

We've decided to modify the scripts so that they are instead run manually by the person running then ensemble. With these changes, the steps to run an ensemble member would be:
1. Clone an experiment that has been prepared in the [https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments](experiments repository)
2. Load the payu module and run 
   ```
   python scripts/setup_historical_restart.py` --output-restart <path to write modified restart>
   ```
5. `payu setup` and `payu run`. 

The `setup_historical_restart.py` copies the `config.yaml:restart` directory to the path specified by `--output-restart`. In the copy, it then sets the start date, inserts the 20PgC initial wood product, and the initial wood thinning data.

The path to the updated restart is then placed into the `config.yaml` file, and a commit describing the modifications is automatically added.

I have also changed the wood thinning user script to apply its changes at the end of the run, rather than at the start. This again ensures that the data in the manifests matches the files the model ends up using.

### Considerations
1. The changes to the ice restart date append the current time to the netCDF `history` attribute and the modifications to the UM restart modify the date creation data in the file. As a result, subsequent runs of the script on the same initial restart file will result in different md5 hashes. 
2. How should we set up the base configuration which will be released to users? Will we specify a restart which already has the modifications made, allowing for a user to just use `payu run`? Or will we specify an unmodified restart and require users to run the ensemble scripts? (To me the first makes sense).



@aidanheerdegen @anton-seaice, would you be happy to take a look at this and let me know if the above workflow makes sense/sounds ok to you!

Once this is merged in we should be able to set up the ensembles

blimlim:
@NortonAlex, I made two small changes to the `insert_external_woodprod_CNP.py` script in e9bc6e21f8c17fceeb8dd8db8d119b6fc6b798d2: 
- Allowing it to write output to the same location as the input (this was mainly to help tidy up some of my code in the calling script)
- Wrapping the processing into the `insert_woodprod()` function, so that it could be imported into the calling script.

If you are happy to take a quick look to sanity check those changes, that would be much appreciated!

anton-seaice:
> ###
> 
>     1. The changes to the ice restart date append the current time to the netCDF `history` attribute and the modifications to the UM restart modify the date creation data in the file. As a result, subsequent runs of the script on the same initial restart file will result in different md5 hashes.

So the atmosphere and ocean changes are deterministic?

There is a `-h` option to turn this off in `ncatted`

anton-seaice:
Should ```--output-restart <path to write modified restart>``` just default to the config folder or scratch ? (We don't expect these to be kept right)

blimlim:
> There is a -h option to turn this off in ncatted

Thanks good idea, I've added this in 

> So the atmosphere and ocean changes are deterministic?

The ocean definitely is (just a change to the ocean_solo.res text file). For the atmosphere I was concerned as the UM files contain a generation date field. Testing it out however, the scripts here don't seem to modify it and so they should also be reproducible.

anton-seaice:
> 2\. Will we specify a restart which already has the modifications made, allowing for a user to just use `payu run`? Or will we specify an unmodified restart and require users to run the ensemble scripts? (To me the first makes sense).

Yep first makes sense

anton-seaice:
Is there a sensible way to record the restart files used before the changes are made ?

e.g. run payu setup and commit manifest/restart.yaml at the start of `scripts/setup_historical_restart.py ` ?

scripts/setup_historical_restart.py
```diff
@@ -0,0 +1,106 @@
...
+    config["restart"] = str(restart_dir)
+    YAML().dump(config, Path("config.yaml"))
+
+
+def commit_config(input_restart, output_restart):
+    """Commit changes to the config.yaml"""
+    repo = git.Repo(".")
+    repo.index.add("config.yaml")
+    msg = (
+        f"Restarts in {input_restart} copied to {output_restart} and modified\n"
```
anton-seaice, line 88: I think ideally we'd have a manifest for `input_restart` 

But we can probably just put that in the two hard basket I guess ?

scripts/setup_historical_restart.py
```diff
@@ -0,0 +1,106 @@
...
+            "initial condition in a historical simulation. This includes:\n"
+            " - Resetting the restart date to 1850-01-01\n"
+            " - Inserting an initial wood product of 20PgC into the atmosphere restart\n"
+            "The initial restart path is taken from the config.yaml, and the output restart path is specified as an argument."
+        ),
+        formatter_class=argparse.RawTextHelpFormatter
+    )
+    parser.add_argument(
+        '--output-restart',
+        help='Path for saving modified output ESM1.6 restart directory.',
```
anton-seaice: It's required that output-restart is an absolute path

See https://github.com/payu-org/payu/issues/762

Not sure if we should do anything in this script or not ?

scripts/restart_1850_cleanup.py
anton-seaice, line 1: Is this file used ?

blimlim: Not used anymore! Removed along with others in d32a92d

scripts/setup_historical_restart.py
blimlim: I've added in a line to resolve the output-restart path in 08ccfcb

blimlim:
> Should --output-restart <path to write modified restart> just default to the config folder or scratch ? (We don't expect these to be kept right)

It now defaults to `archive/initial_restart`, allowing the user to run with:
1. `python scripts/setup_historical_restart.py`
2. `payu run`

As @anton-seaice has brought up, one risk with this is if the user runs `payu sweep --hard`. In this case, they will need to regenerate the restart.

scripts/setup_historical_restart.py
```diff
@@ -0,0 +1,131 @@
...
+import insert_external_woodprod_CNP
+import update_thinning
+
+# Filepaths required for the initial wood product script
+EXTERNAL_NC_PATH = "/g/data/vk83/prerelease/configurations/inputs/access-esm1p6/modern/historical/atmosphere/land/biogeochemistry/global.N96/2026.05.08/access-esm16_wood_prod_initial_1850_20PgC.nc"
+STASHMASTER_BASE_PATH = "/g/data/vk83/prerelease/configurations/inputs/access-esm1p6/share/atmosphere/stash/2026.01.21/STASHmaster/STASHmaster_A"
+STASHMASTER_EXT_PATH = "./atmosphere/prefix.PRESM_A"
+
+# Filepaths required for wood thinning
+THINNING_FILE = "/g/data/vk83/configurations/inputs/access-esm1p6/modern/historical/atmosphere/land/vegetation/global.N96/2026.05.08/LUH3_cable_thinning_frac_from_bioh_1850-2023_v2.nc"
```
blimlim, line 19: I'm not super happy about hard coding the input file paths here, as it risks this script falling out of sync with the config.yaml. However I haven't come up with any better approach

scripts/setup_historical_restart.py
anton-seaice: @aidanheerdegen - the workflow here is that `restart:` path is set by the user, then the restart files are copied by this script, modified and then the `restart:` field is updated to the modified files.

So ideally there would be a manifest before the changes are made by this script ?

aidanheerdegen: It would be possible to follow the logic in the [`payu.manifest.Manifest` class](https://github.com/payu-org/payu/blob/master/payu/manifest.py#L283-L298) but for just the restart manifest, and use `calculate_fast` to update the `binhash` checksums but use the MD5 checksums from the previous restart manifest

https://github.com/payu-org/payu/blob/master/payu/manifest.py#L59

scripts/setup_historical_restart.py
anton-seaice: Shall we make an issue for this ?

blimlim:
I've done a 5 year test run with the current version of the scripts. It started with restart `/g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/Ndep2-PI-CNP-emissions/restart199/`, the restart used by the first esm-historical run in `/g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/esm-historical-01` minus the modifications. 

At the end of 5 years, model output is identical to the esm-historical run: 

```
> nccmp -dg aiihca.pa-185412_mon.nc /g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/esm-historical-01/output004/atmosphere/netCDF/aiihca.pa-185412_mon.nc
>
> nccmp -dg iceh-1monthly-mean_1854-12.nc /g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/esm-historical-01/output004/ice/iceh-1monthly-mean_1854-12.nc 
>
```

blimlim:
For the default restart in the configuration, we'll use one that already has the modifications applied so that a user can clone the configuration and run it, without needing to run the restart modification scripts.

The current default restart `/g/data/vk83/configurations/inputs/access-esm1p6/modern/historical-emissions/restart/2026.05.27` has the date and initial wood product modifications, but no wood thinning modifications as the original version of the wood thinning script applied the changes at the start of the run. However, this PR changes the [thinning script](https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/601-historical-ensemble-scripts/scripts/update_thinning.py) to run at the end of the run. Hence we need to update the default restart with the initial wood thinning data.


Also noting: For generating the ensembles, it will be necessary for the person running them to run the restart modification scripts.

anton-seaice:
Gave this a run, seems to set the dates correctly in the log files and output

anton-seaice (dismissed):

blimlim:
!test repro commit

blimlim:
!test repro commit

anton-seaice (dismissed):

scripts/setup_historical_restart.py
blimlim: Added one for this here: https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/819

scripts/setup_historical_restart.py
blimlim: I think we've decided to leave this change out

anton-seaice (approved):

scripts/set_restart_1850.sh
blimlim: We ended up copying the updated script from https://github.com/ACCESS-NRI/umfile_utils/blob/main/src/umfile_utils/change_dump_date.py to here

Issue #746, opened 2026-05-15
Initial `release-` branches


anton-seaice:
For each configuration, a commit is proposed to use for an initial `release-` branch based on what is being used for spinup experiments

| config/experiment | config commit | experiment branch |
|----------------|--------------|---------------|
| piControl | 946134fbad251a67ff7c6ca5f76c3cfb9c95d707 | https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments/tree/dev-preindustrial%2Bconcentrations-22-04-26/
| esm-piControl| 2e3530774c6f8a0fe9e92f6ddceb132f219fedf5 | https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments/tree/dev-preindustrial%2Bemissions-14-04-26 |
| esm-flat10 | e3095cb9158803996276ac8ccd9b72de7acbfc5c | https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments/tree/dev-flat10-29-04-26/ |
| 1pctCO2 | 7f5b6fab09b6611d6a2fd7e68a6277becbd2f6b6 | https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments/tree/dev-1pctCO2-15-04-26 | 
| 1pctCO2-bgc | 4bf5f99264cbbe6dd3a54b820371ceffc5d9bba8 | https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments/tree/dev-1pctCO2-bgc-15-04-26 | 
| 1pctCO2-rad | d34da474ed24aef8c17a62da95011aa5301fcbe3 | https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments/tree/dev-1pctCO2-rad-15-04-26 |
| abruput-4xCO2 | 9670a06b9e8bfe47852fcb7ab633c2ecd90e11c0 | https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments/tree/dev-4xCO2%2Bconcentrations-16-04-26 |

Based on https://forum.access-hive.org.au/t/status-of-access-esm1-6-experiments/6337, i believe they are the experiments which have had a final run started ? @JhanSrbinovsky @rml599gh - if you can sanity check the table that would be great

[...]

PR #764, opened 2026-05-19
Use release versions for CI on `release-` branches

anton-seaice:
Contributes to #746 

use release versions of payu and model-config-tests for `release-` branhces

### DIFF
--- a/config/ci.json
+++ b/config/ci.json
@@ -26,6 +26,10 @@
     "qa": {
         "default": {
             "markers": "access_esm1p6 or config"
+        },
+         "release-.*": {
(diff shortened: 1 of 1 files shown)


config/ci.json
blimlim, line 28: Should we be running the qa tests on the release branches, or not yet?

anton-seaice: I think this is already is ?

But ill set the versions explicity for qa too

blimlim: > I think this is already is ?

Oh wait yep, I think you're right

blimlim (approved):
Thanks @anton-seaice, looks good

Issue #733, opened 2026-05-05
Remove concentration STASHC from emissions driven configurations


blimlim:
An earlier cherry-pick accidentally copied the `_concentrations` STASHC files into the dev-historical+emissions configuration. These can be removed as they aren't used for that configuration.

Issue #731, opened 2026-05-05
Remove volcts and solar constant from coupling namelist in historical configuration


blimlim:
The historical configurations include constant values for the volcanic forcing and solar constant specified in the `atmosphere/input_atm.nml` namelist:
https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/b9dbd0ddbbeabd8dc367c169062f11c0959b122a/atmosphere/input_atm.nml#L4-L5

Since these configurations read in time varying values [from files](https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/b9dbd0ddbbeabd8dc367c169062f11c0959b122a/atmosphere/namelists#L906-L909) for these quantities, the constant values in `input_atm.nml` aren't used. These could be removed to avoid confusion.

Issue #730, opened 2026-05-05
Remove unused input file from historical configurations


blimlim:
The dev-historical+concentrations/emissions configuration's config.yaml files still reference an input file from CMIP6 which is no longer used: https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/b9dbd0ddbbeabd8dc367c169062f11c0959b122a/config.yaml#L49

This should be removed.

Issue #23, opened 2025-01-28
Redundant ancillary file paths in ihist.


MartinDix:
Some of the entries in `ihist` like
```
OROG    = 'OROG : $ANCIL_LAND/qrparm.orog
```
aren't used because the orography is not updated by the model.  

In other cases like
```
OZONE   = 'OZONE : $CMIP5ANCIL_AEROSOL/ozone_1850_ESM1.anc   
```
the explicit path is ignored and the value from the environment variable in `um_env.yaml` is used
```
OZONE: INPUT/ozone_1850_ESM1.anc
```

The unused variables like `OROG` should be set to empty strings (need to check whether they can actually be deleted). Others should have the old unused path removed.

Applies to both ESM1.5 and 1.6 configurations

MartinDix:
The `ihist` namelists can be appended with those in `namelists` with a change in um_en.yml to set `UNIT11: namelists`.

The ancillary file paths could then be set directly there rather than in `um_env.yml`, though I think it's clearer to leave them as they are.

All the variables set in the `NLCFILES` namelist have been initialised to an empty string in the model via the equivalenced `model_ft_unit` variable 
https://github.com/ACCESS-NRI/UM7/blob/ad219516fcbb5386739e3e07ab7fef6a54634034/umbase_hg3/src/control/top_level/initchst.F90#L122
so the blank and unused ones can be deleted from the namelist.

blimlim:
Configurations contain inconsistencies in the`ihist` outside of the `NLCFILES` section. The following diff compares the piControl and amip configurations:
```diff
- &NLIHISTO
- MODEL_DATA_TIME = 6*-32768,
- RUN_MEANCTL_RESTART     =                     0,
- RUN_INDIC_OP    =                     0,
- RUN_RESUBMIT_TARGET     = 6*-32768,
- FT_LASTFIELD    = 142*-32768
+ &NLIHISTO 
  /
+
  &NLCHISTO
- RUN_HIST_TYPE   = 'Interim   ',
- RUN_TYPE        = 'Reconfig',
- RUN_COMPCODE    = '              ',
- RUN_LAST_MEAN   = '              ',
- RUN_MEANS_TO_DO = ' ',
- RUN_OCEAN_FIRST = ' ',
- RUN_JOB_NAME    = 'aiihc000',
- RUN_ID  = '     ',
- RUN_RESUBMIT    = 'Y',
[...]

PR #765, opened 2026-05-20
730 731+historical clean up

blimlim:
**1. Summary**:

This PR cleans the following parts of the esm-historical configuration:
- Remove STASHC files which are only relevant for the concentration driven configurations
- Remove the constant values for the volcanic forcing and solar constant in `input_atm.nml`, as time varying values are instead read from the files specified here:  https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/d124dd1a307dc5c3ea5ecf5aeabf70658f871326/atmosphere/namelists#L907-L909
   A potential side effect: If the model is run outside of years 1700-2300, it [will fall back](https://github.com/ACCESS-NRI/UM7/blob/d47b0b5b91334dacf69fc91a493138f05d901893/umbase_hg3/src/control/top_level/glue_rad-rad_ctl2.F90#L2349-L2362) to using `volcts_val` and its [default of 97.33]( https://github.com/ACCESS-NRI/UM7/blob/d47b0b5b91334dacf69fc91a493138f05d901893/umbase_hg3/src/control/top_level/volcts_Mod.F90#L36)  if it's not specified in the namelist. This shouldn't happen in this configuration as the ancillaries are only valid for 1850-2022. However, we can omit this change if we think it could cause issues. 
- Removes the old `Bio_1849_2015_ESM1.anc` file carried over from ESM1.5 which is not used in this configuration
- Removes all filepaths from the `ihist` file. These don't appear to be used and can lead to confusion as some of these paths are instead set in the `um_env.yaml` file. Once this change is copied to all the configurations it should also simplify the comparisons between configurations.

**2. Issues Addressed:**

- #730 
- #731 
- #23 
- #733 

- [ ] input workflow:

**4. Ad-hoc Testing**

Comparing runs with and without these changes, the atmosphere outputs are identical after 5 years:

```
> nccmp -dgf real-base/archive/output004/atmosphere/netCDF/aiihca.pa-185412_mon.nc all-combined//archive/output004/atmosphere/netCDF/aiihca.pa-185412_mon.nc
> 
```
I'll keep the runs going for 5 years and add an update when complete.

**5. CI Testing**

**6. Reproducibility**

- [x] Yes

**7. Performance**

- [ ] Yes
- [ ] No

These changes only tidy up the configuration and don't affect how the model runs.

**8. Manifests**

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

```yaml
manifest:
  reproduce:
    exe: false
    input: false
    restart: false
runlog:
  enable: false
```

**9. Documentation**

Is the documentation updated?
- [ ] Yes
- [x] N/A

**10. Merge Strategy**

- [x] Merge commit
- [ ] Rebase and merge
- [ ] Squash


blimlim:
!test repro

blimlim:
Testing with a longer run, results are identical to the production run (`/g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/esm-historical-01`) after 10 years:

```
> nccmp -dg aiihca.pa-185912_mon.nc /g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/esm-historical-01/output009/atmosphere/netCDF/aiihca.pa-185912_mon.nc
> 
> nccmp -dg iceh-1monthly-mean_1859-12.nc /g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/esm-historical-01/output009/ice/iceh-1monthly-mean_1859-12.nc
>
```

MartinDix (approved):
Looks good.

Issue #827, opened 2026-06-16
Ensemble scripts for amip configuration


blimlim:
An amip ensemble will be run members branching from different historical simulations in 1978. For each ensemble member, the ocean, sea ice, and coupling files need to be removed from the initial restart, and coupling related fields will need to be removed from the atmosphere restart file. This can be done following step 2 in [these instructions](https://forum.access-hive.org.au/t/create-an-amip-restart-from-a-coupled-restart/5261). The restart date will not need to be changed, as it will already use 1978 as the year.

To simplify the setup and running of the ensemble, it would be good to add scripts to the amip configuration which apply these modifications to a selected historical restart. These could work in a similar way to the [scripts in the historical configurations](https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/dev-historical/scripts/setup_historical_restart.py), which set up a restart in the archive directory and update the config.yaml.

For the um_fields_subset.py script, we could copy the updated version from  https://github.com/ACCESS-NRI/umfile_utils/blob/main/src/umfile_utils/um_fields_subset.py into the configuration.

PR #843, opened 2026-07-01
Add script to setup amip restart from 1978 historical restart

blimlim:
**1. Summary**:

AMIP experiments branch off the 1978 restart files produced by historical experiments. Before running, the initial restart needs to be modified for compatibility with the AMIP configuration:
- Ocean, coupler, and sea ice files can be removed. This is might not be 100% necessary but it does simplify the initial restart manifest.
- Coupling related fields need to be removed from the atmosphere restart, following the process in step 2 [here](https://forum.access-hive.org.au/t/create-an-amip-restart-from-a-coupled-restart/5261)

Following discussions, we decided to follow the approach from the historical and esm-historical configurations for modifying the initial restart. Those configurations contain a script which can be run by the user, which:
- copy the restart specified in the config.yaml to the archive directory
- apply modifications to the copied restart
- change the path in the config.yaml to point to the modified restart, and commit the change to the runlogs with a descriptive message.
The default restart in those configurations already contain the modifications, and so the script is only necessary when using a different restart.

This PR add s a script which performs the above steps to the amip configuration.

**2. Issues Addressed:**

- #827

- [ ] input workflow:

**4. Ad-hoc Testing**

Comparing two one year runs. The first used the scripts to modify the initial restart, which was first copied from the historical simulation. For the second, I manually modified the restart following step 2 [here](https://forum.access-hive.org.au/t/create-an-amip-restart-from-a-coupled-restart/5261).

Outputs are identical at the end of the year:
```
nccmp -dgf manual-restart-comparison/archive/output000/atmosphere/netCDF/aiihca.pa-197812_mon.nc update-um_fields_subset//archive/output000/atmosphere/netCDF/aiihca.pa-197812_mon.nc 
```

**5. CI Testing**

**6. Reproducibility**

- [ ] Yes

**7. Performance**

- [ ] Yes
- [ ] No

**8. Manifests**

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

```yaml
manifest:
  reproduce:
    exe: false
    input: false
    restart: false
runlog:
  enable: false
```

**9. Documentation**

Is the documentation updated?
- [ ] Yes
- [ ] N/A

**10. Merge Strategy**

- [ ] Merge commit
- [ ] Rebase and merge
- [ ] Squash

### DIFF
--- a/scripts/um_fields_subset.py
+++ b/scripts/um_fields_subset.py
@@ -15,7 +15,11 @@ import warnings
 from textwrap import dedent
 from itertools import chain
 
-PROGNOSTIC_STASH_CODES = tuple(chain(range(1,999+1), range(33001,34999+1)))
+# Prognostic variables have section 0, 33 (tracers), or 34 (UKCA).
+# Tracer flux variables 3100 - 3129 are also treated as prognostic.
+PROGNOSTIC_STASH_CODES = tuple(chain(range(1,999+1), range(3100,3129+1), range(33001,34999+1), range(54000,54999+1)))
+# Tracer variables have section 33
+TRACER_STASH_CODES = tuple(range(33000, 33999+1))
 
 def convert_to_list(value: str):
     """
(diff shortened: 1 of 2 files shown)


blimlim:
!test repro

blimlim:
!test repro

blimlim:
!test repro

blimlim:
!test repro

blimlim:
!test repro

scripts/um_fields_subset.py
```diff
@@ -0,0 +1,351 @@
+#!/usr/bin/env python
```
blimlim, line 1: This script is copied directly from https://github.com/ACCESS-NRI/umfile_utils commit e02b84c827964e43f2c7107f2b5e88b9580c0b68

NortonAlex:
Hi @blimlim, not sure how much I can help here. Would be best to let Martin review it as well. 

One question though. Will the "exclude_list" specified in scripts/setup_amip_restart.py always be the same? Or should "prog_fields_mismatch.py" be run as part of this workflow so that the exclude_list is determined for each run?

blimlim:
Thanks @NortonAlex for taking a look. 
> One question though. Will the "exclude_list" specified in scripts/setup_amip_restart.py always be the same? 

Good question. I think it will be sufficient to hard code the list of fields for the cases where this script will be run. From what I understand, the list shouldn't change unless the UM is modified to add new coupling fields. We should be safe from this at least for the current version of the model

MartinDix:
At the moment the payu python includes mule because of um2nc. Need to ensure that it still does when we move to a stand-alone um2nc.

MartinDix (approved):
Looks good

Issue #890, opened 2026-07-28
Archive configs for idealised experiments


anton-seaice:
For the idealised experiments, which are not planned for long term support, these can be archived to an `archive-` tag and the config docs updated (https://access-esm1p6-configs.access-hive.org.au/#repository-structure) 

Configs impacted:

`abrupt-4xCO2`
`1pctCO2` + variants
`flat10` + variants

Issue #871, opened 2026-07-17
Update scheduled tests markers


anton-seaice:
There's currently a slightly random set of configs getting tested for scheduled tests:

https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/7ef10f49f68308953c4bf11280402892ac07cc46/config/ci.json#L7-L15

We should probably include some `release-` branches, and clarify what the justification is for the selection that is there.

anton-seaice:
Closed in #892

PR #892, opened 2026-07-28
Add release configurations to scheduled testing

anton-seaice:
Update the scheduled testing to reflect current branches per #890  and include some release branches per #871


blimlim (approved):
Thanks @anton-seaice, looks good

PR #935, opened 2026-08-06
Refactor thinning update file time indexing

NortonAlex:
Updated wood thinning year selection to use the NetCDF time coordinate rather than assuming an 1850 reference year, which would fail for future scenario run that uses an input file with first time index of 2022. The script now decodes CF-compliant time coordinates, matches the restart file t2_year against the thinning dataset year values, determines the corresponding time index, and raises an error if the requested year is not present. This allows thinning datasets with arbitrary start years (e.g. 2022-2100) to be used correctly.

**1. Summary**:

What has changed?
Fixed year selection in update_thinning.py when reading thinning data from an external NetCDF file. Rather than assuming the thinning dataset starts in 1850 and computing `t_index = model_year - 1850`. The script now matches the restart year (t2_year) to the years present in the NetCDF time coordinate to determine the corresponding time index.

Why was this done?

The previous implementation only worked for datasets beginning in 1850 with one timestep per year. New scenario input thinning datasets begin in 2022, so the calculated time index could be incorrect and select the wrong thinning year (or an out-of-range index). 

**2. Issues Addressed:**

-

- [ ] input workflow:

**4. Ad-hoc Testing**

Tested on a H-scenario thinning dataset spanning 2022-2100. Confirmed that the restart year (t2_year) is correctly matched to the corresponding year in the NetCDF time coordinate and that the resulting time index (t_index=0) is passed to replace_field(). Added validation to raise an informative error if the requested year is not present in the thinning dataset.

**5. CI Testing**

**6. Reproducibility**

- [ ] Yes
- [x] No - `!test repro commit` has been run. 

This change is expected to alter results when thinning datasets do not begin in 1850 e.g. for future scenario runs. The previous behaviour could select an incorrect thinning year or fail if t_index was outside range of input thinning NetCDF file time coordinate. The new behaviour selects the year matching the restart t2_year.


blimlim (approved):
Thanks @NortonAlex for putting this in. The changes look good, and testing with historical restarts gives identical results with and without the updates

NortonAlex:
Hey @blimlim, I have just cloned this branch and tested it. I get an xarray error when running `payu setup`. I presume this is an issue with the version of xarray that is being used, it's hard for me to tell which exact version is getting used. 

`
Traceback (most recent call last):
  File "/g/data/p66/ajn563/ACCESS-ESM/ESM1.6/test-scen7-h-rampnewtiles/./scripts/update_thinning.py", line 39, in <module>
    time_coder = xarray.coders.CFDatetimeCoder(use_cftime=True)
AttributeError: module 'xarray' has no attribute 'coders'
[ERROR] User defined script/command failed to run: ./scripts/update_thinning_driver.sh
Error: Command '['./scripts/update_thinning_driver.sh']' returned non-zero exit status 1.
`

blimlim:
I think this is an issue comming from this line:
https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/39290dc37407eb731bb9b9880ba8f9e23e412552/scripts/update_thinning.py#L1

which ends up causing the script to use an older python environment. This has recently been fixed in the `dev-historical` branch. @NortonAlex, would you be happy for me to rebase `test-scen7-h` onto `dev-historical` to bring in the more recent updates?

NortonAlex:
Yes, that would be great and I think needed soon anyway. Thanks

blimlim:
I've rebased the `test-scen7-h` and `test-scen7-vl` branches and so hopefully this one should work now!

NortonAlex:
Super. Thanks

Issue #313, opened 2025-12-09
Integration with data specifications


joshuatorrance:
The data specifications are close to done and we'll need to figure out how to get them integrated with the config, payu, addmeta, etc.

- [Schema](https://github.com/ACCESS-NRI/schema/tree/main/au.org.access-nri/model/output/file-metadata/2-0-0) (some pending updates to schema in [this](https://github.com/ACCESS-NRI/schema/pull/58) PR)
- [Docs](https://access-output-data-specifications--2.org.readthedocs.build/en/2/specification/)

Thoughts on how to proceed?
@aidanheerdegen @blimlim

joshuatorrance:
Great! When I find the time I'll grab a copy of that directory and do a alpha test run of addmeta with the new specification.

joshuatorrance:
I finally got around to scripting the proof of concept test of `addmeta` with a copy of the suggested dataset. The script I used can be found here: `/scratch/tm70/jt4085/esm16_examples/addmeta_script.sh`

With the way I did things it would be useful to have an extra feature in `addmeta` where we can add a file of template vars that can be used to populate the jinja keys used to flesh out the attributes. Maybe allowing template vars to be specified directly on the command line too? @aidanheerdegen - Do we have a better way to handle this?

The filenames could be better engineered to make life easy for `addmeta`:
- The `frequency` portions of the current filenames are not consistent between realms and don't follow the expected vocab (i.e. `1mon`, `1day`, `1yr`, ...). The static/fixed filenames also don't have a frequency portion which means `addmeta` can't extract it, `addmeta` needs those filenames to have `fx` as their frequency.
- We can't do `variable_id` for everything yet but that obviously needs to wait for the one variable per file stuff to be finalised.

There were also a couple of attributes I didn't spot in the various .yaml files (`base_configuration` & `run_id`) but I presume payu would be able to find them.

aidanheerdegen:
Thanks for trying this anger @joshuatorrance.

> With the way I did things it would be useful to have an extra feature in `addmeta` where we can add a file of template vars that can be used to populate the jinja keys used to flesh out the attributes. Maybe allowing template vars to be specified directly on the command line too? [@aidanheerdegen](https://github.com/aidanheerdegen) - Do we have a better way to handle this?

Can you be more explicit about which values you're referring to? Is it the realm for each submodel?

[...]

PR #950, opened 2026-08-12
Add data spec changes

blimlim:
**1. Summary**:
This draft PR adds the ACCESS-NRI data spec changes to the configurations, and is partly an update to the older PR  #523. Changes include:
- Ocean filenames are changed to match the new conventions. This is partly done in the diag_tables, and is finished using the rename_ocean.py script (based on that in https://github.com/ACCESS-NRI/esm1.6-scripts/pull/14).
- Atmosphere and ice changed to use yearly output files rather than monthly
- `splitnc` is called in the post-processing script to split the ice outputs into single variable files
- `um2nc` uses the new esm1p6 driver, and writes output to single variable files.
- `addmeta` is called as the last step in the post script.

This PR is still a draft, as: 
- It's using a staging model-processing module
- There are still Fixme's in the addmeta files

**2. Issues Addressed:**

- #313

- [ ] input workflow:

**4. Ad-hoc Testing**

**5. CI Testing**

**6. Reproducibility**

- [ ] Yes

**7. Performance**

- [ ] Yes
- [ ] No

**8. Manifests**

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

```yaml
manifest:
  reproduce:
    exe: false
    input: false
    restart: false
runlog:
  enable: false
```

**9. Documentation**

Is the documentation updated?
- [ ] Yes
- [ ] N/A

**10. Merge Strategy**

- [ ] Merge commit
- [ ] Rebase and merge
- [ ] Squash


scripts/post-processing/post_process.sh
```diff
@@ -0,0 +1,80 @@
...
+# Convert UM atmosphere fields files to netCDF.
+# By default UM atmosphere fields files are deleted after conversion to save space. 
+# Remove --delete-ff command line option to retain original files for testing purposes
+# Source code for the um2nc command is available at https://github.com/ACCESS-NRI/um2nc-standalone/
+um2nc driver esm1p6 $PAYU_CURRENT_OUTPUT_DIR --delete-ff  --one-nc-per-stash-variable
+
+
+# Split sea ice output into single variable files using the splitnc command. The splitnc command calls https://github.com/ACCESS-NRI/splitnc/blob/main/src/splitnc/splitnc.py
+icefiles=($PAYU_CURRENT_OUTPUT_DIR/ice/iceh-*)
+if [ "${#icefiles[@]}" -gt 0 ]
```
blimlim, line 26: This is included to prevent issues with reruns of the script if it fails on the first attempt

blimlim:
!test repro

scripts/post-processing/addmeta/dataspec.yaml
```diff
@@ -0,0 +1,64 @@
...
+    id: "{{ metadata.experiment_uuid }}"
+    license: "{{ metadata.license }}"
+    model: "{{ metadata.model }}"
+    model_version: "{{ metadata.version }}"
+    # FIXME: we may need a semi-manual mapping for branch time one
+    parent_experiment_branch_time: "0001-01-01T00:00:00Z"
+    # FIXME: parent_experiment is missing in the test dataset
+    parent_experiment_uuid: "45fa297e-165c-4944-98a6-dbb653be30fc" #"{{ metadata.parent_experiment }}"
+    # FIXME: PAYU_RUN_ID is missing from my test env.yaml file
+    run_id: "{{ env.RUNID }}"
```
blimlim: Payu now exports `PAYU_RUN_ID` which I think we can use here

scripts/post-processing/addmeta/dataspec.yaml
```diff
@@ -0,0 +1,64 @@
...
+    geospatial_lon_min: 180
+    grid: "access-1deg-global"
+    id: "{{ metadata.experiment_uuid }}"
+    license: "{{ metadata.license }}"
+    model: "{{ metadata.model }}"
+    model_version: "{{ metadata.version }}"
+    # FIXME: we may need a semi-manual mapping for branch time one
+    parent_experiment_branch_time: "0001-01-01T00:00:00Z"
+    # FIXME: parent_experiment is missing in the test dataset
+    parent_experiment_uuid: "45fa297e-165c-4944-98a6-dbb653be30fc" #"{{ metadata.parent_experiment }}"
```
blimlim: I guess we grab this from `metadata.parent_experiment` if it exists?

scripts/post-processing/addmeta/dataspec.yaml
```diff
@@ -0,0 +1,64 @@
...
+    geospatial_lat_min: -90
+    geospatial_lon_max: -180
+    geospatial_lon_min: 180
+    grid: "access-1deg-global"
+    id: "{{ metadata.experiment_uuid }}"
+    license: "{{ metadata.license }}"
+    model: "{{ metadata.model }}"
+    model_version: "{{ metadata.version }}"
+    # FIXME: we may need a semi-manual mapping for branch time one
+    parent_experiment_branch_time: "0001-01-01T00:00:00Z"
```
aidanheerdegen:
```suggestion
    parent_experiment_branch_time: "{{ metadata.parent_experiment_branch_time }}"
```
`parent_experiment_branch_time` was added in https://github.com/payu-org/payu/pull/811. 

It will need to be added to existing experiments that do not include this automatically.

scripts/post-processing/addmeta/dataspec.yaml
aidanheerdegen: Yep.

scripts/post-processing/addmeta/dataspec.yaml
aidanheerdegen: Yep.

scripts/post-processing/post_process.sh
```diff
@@ -0,0 +1,80 @@
...
+#PBS -l jobfs=0GB
+#PBS -q normal
+#PBS -l walltime=02:30:00
+#PBS -l wd
+#PBS -W umask=027
+
+set -e
+
+module use /g/data/vk83/staging/modules
+module load model-processing/20260805T125219-7abd909-pr90
```
aidanheerdegen: I know this was flagged in the PR description, but just so it doesn't get accidentally forgotten, this needs to be updated to a released environment.

scripts/post-processing/post_process.sh
aidanheerdegen: Maybe add that as a comment in the code?

scripts/post-processing/post_process.sh
```diff
@@ -0,0 +1,80 @@
...
+
+
+# Split sea ice output into single variable files using the splitnc command. The splitnc command calls https://github.com/ACCESS-NRI/splitnc/blob/main/src/splitnc/splitnc.py
+icefiles=($PAYU_CURRENT_OUTPUT_DIR/ice/iceh-*)
+if [ "${#icefiles[@]}" -gt 0 ]
+then
+    splitnc --shared-vars uarea,tmask,tarea --excluded-vars VGRD.  --use-esm1p6-filenames --fix-cell-methods --overwrite "${icefiles[@]}"
+
+    # Remove the original sea ice model output:
+    rm "${icefiles[@]}"
```
aidanheerdegen:
```suggestion
    if [ $? -eq 0 ]; then
      rm "${icefiles[@]}"
    fi
```
Current code is safe while `set -e` is part of the script. If this were commented out or removed then the original files could be deleted even if `splitnc` failed. I suggest putting a guard around this potentially catastrophic case.

scripts/post-processing/post_process.sh
```diff
@@ -0,0 +1,80 @@
...
+    $PAYU_CURRENT_OUTPUT_DIR/ocean/ocean.*.nc
+
+addmeta \
+    -v -s \
+    -m scripts/post-processing/addmeta/realm_only.yaml \
+    --datavar freq='ocnBgchem' \
+    $PAYU_CURRENT_OUTPUT_DIR/ocean/ocnBgchem.*.nc
+
+# Rename ocean files to remove the realm and "_" in the date
+python3 scripts/post-processing/rename_ocean.py $PAYU_CURRENT_OUTPUT_DIR/ocean/{ocean,ocnBgchem}.*.nc
```
aidanheerdegen: I really don't like that we are locking ourselves into *always* changing ocean file names. Is there really no other option?

scripts/post-processing/addmeta/ocean.yaml
```diff
@@ -0,0 +1,2 @@
+# Intentionally left empty for ocean
+global:
```
aidanheerdegen, line 1: If it's empty then it isn't required.

aidanheerdegen (commented):
I'm nervous that there is no testing, especially with regexes, which are close to impossible to grok effectively. Not sure I have a suggestion about how to do this apart from making a separate package.

scripts/post-processing/addmeta/ocean.yaml
joshuatorrance: That's just an artefact from my [cylc suite](https://github.com/ACCESS-NRI/esm1.6-scripts/blob/a17f0895c5872c9bdfa2743fbc985af9ff9628ff/esm1p6_postprocessor/flow.cylc#L139) where I was using `{{ realm }}.yaml` and there I preferred an empty file to extra logic.

scripts/post-processing/addmeta/realm_only.yaml
joshuatorrance, line 1: This was used in the [cylc suite](https://github.com/ACCESS-NRI/esm1.6-scripts/blob/a17f0895c5872c9bdfa2743fbc985af9ff9628ff/esm1p6_postprocessor/flow.cylc#L230) to add `realm` to the ocean/oceabgc files. I doubt we need it here.

edit: Having scrolled down and read the rest, perhaps we do need it after all! I'd misinterpreted the `realm: ocnBgchem` changes in the ocean yamls above.

scripts/post-processing/post_process.sh
```diff
@@ -0,0 +1,80 @@
...
+if [ "${#icefiles[@]}" -gt 0 ]
+then
+    splitnc --shared-vars uarea,tmask,tarea --excluded-vars VGRD.  --use-esm1p6-filenames --fix-cell-methods --overwrite "${icefiles[@]}"
+
+    # Remove the original sea ice model output:
+    rm "${icefiles[@]}"
+fi
+
+
+# Clean up global metatdata to meet ACCESS-NRI dataspec standards
```
joshuatorrance, line 35: I think this section can be tidied up a bit? Something like:
```bash
# Add realm to the ocean files before they're renamed
addmeta \
    -v -s \
    -m scripts/post-processing/addmeta/realm_only.yaml \
    --datavar realm='ocean' \
    $PAYU_CURRENT_OUTPUT_DIR/ocean/ocean.*.nc

addmeta \
    -v -s \
    -m scripts/post-processing/addmeta/realm_only.yaml \
    --datavar freq='ocnBgchem' \
    $PAYU_CURRENT_OUTPUT_DIR/ocean/ocnBgchem.*.nc

# Rename ocean files to remove the realm and "_" in the date
python3 scripts/post-processing/rename_ocean.py $PAYU_CURRENT_OUTPUT_DIR/ocean/{ocean,ocnBgchem}.*.nc

# Add the required metadata to all the .nc files
for realm in "atmosphere ice ocean"; do
    addmeta \
      -v -s \
      -d metadata.yaml \
      -d $PAYU_CURRENT_OUTPUT_DIR/env.yaml \
      -m scripts/post-processing/addmeta/dataspec.yaml \
      -m scripts/post-processing/addmeta/$realm.yaml \
      --fnregex='access-esm1p6\.\w+(?:\.\dd)?\.(?P<var>\w+)\.(?P<freq>\w{2,4})(?:\.\w+)?(?:\.\d{4})?\.nc' \
      $PAYU_CURRENT_OUTPUT_DIR/$realm/*.nc
done
```

We'd need the "empty" `realm.yaml` files for this to work as written. There are some other options too:
- could build the command as a string, add the `realm` option if needed and the execute the string
- could add the realm to `ice` files in a separate invocation of `addmeta` then get rid of the loop and just have a `*` for the subdirectory. Costs an extra file touch of course.

scripts/post-processing/post_process.sh
joshuatorrance: What happens if the script fails or is interrupted part way through running `splitnc` on the ice files? If I'm reading it right it won't rerun `splitnc`?

scripts/post-processing/post_process.sh
```diff
@@ -0,0 +1,80 @@
...
+# Remove --delete-ff command line option to retain original files for testing purposes
+# Source code for the um2nc command is available at https://github.com/ACCESS-NRI/um2nc-standalone/
+um2nc driver esm1p6 $PAYU_CURRENT_OUTPUT_DIR --delete-ff  --one-nc-per-stash-variable
+
+
+# Split sea ice output into single variable files using the splitnc command. The splitnc command calls https://github.com/ACCESS-NRI/splitnc/blob/main/src/splitnc/splitnc.py
+icefiles=($PAYU_CURRENT_OUTPUT_DIR/ice/iceh-*)
+if [ "${#icefiles[@]}" -gt 0 ]
+then
+    splitnc --shared-vars uarea,tmask,tarea --excluded-vars VGRD.  --use-esm1p6-filenames --fix-cell-methods --overwrite "${icefiles[@]}"
```
joshuatorrance: I'd recommend using [`--skip-existing`](https://github.com/ACCESS-NRI/splitnc/blob/b1563de8bf25b7124d4c47965abdad3dcf07a2c1/src/splitnc/splitnc.py#L663) rather than `--overwrite` here.

I still need to document that new feature properly but in brief:
- for all modes `splitnc` now writes output to `outfile.nc.temp`, then moves that files to `outfile.nc` thus if `to_netcdf` is interrupted we don't end up with a bad `outfile.nc`
- `--skip-existing` will skip writing out to `outfile.nc` if it already exists, thus speeding up `splitnc` on reruns

joshuatorrance: Do we need to group monthly output into years here? If so we'll need something like:
`--input-group-regex 'iceh-1(daily|monthly)-mean_\d{4}-(?P<wild>\d{2})\.nc'`

blimlim: Ah good point! I think you'd previously mentioned to me about the temporary output locations but I'd forgotten. I'll swap over to `--skip-existing`.

This PR also changes the ice model to output yearly files, and so we only need to split them without concatenating.

scripts/post-processing/addmeta/dataspec.yaml
blimlim: Ah fantastic!

scripts/post-processing/post_process.sh
blimlim: The original ice files will only be deleted on successful completion of the `splitnc` command, and so if it fails part way through, `splitnc` should still be called on the rerun.

`splitnc` produces an error if it doesn't find any input files, and so this check should allow for reruns if something fails after `splitnc`

scripts/post-processing/post_process.sh
blimlim: Very good point, thanks!

scripts/post-processing/post_process.sh
blimlim: @aidanheerdegen, I understand it's fairly messy this way. Were there additional concerns/problems the renaming could lead to?

The alternative I can think of would be to maintain a list (maybe in a yaml file) of the MOM variables which should have realm set to `ocnBgchem`, and feed this into addmeta. I think this would be < 100 variables, and unlikely to change over time.

Another alternative would be to try and extract the information from the [`diag_table_..._source.yaml`](https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/dev-piControl/ocean/diagnostic_profiles/source_yaml_files/diag_table_CMIP7_core_source.yaml) files, e.g. we could specify the realm at the top of each section. I'm not sure whether that could end up a bit fragile though, e.g. we don't actually know which source.yaml file might correspond to the actual `diag_table` used by a user


Would either of the above sound a bit better?

aidanheerdegen: My objection is we end up locking every ocean model into a post-processing step that otherwise not be necessary once we have `addmeta` incorporated into `payu`.

Yeah @blimlim and I had a brief discussion about extracting the info from the diag table source yaml file, but @blimlim pointed out that it was tricky to know which file had produced which `drag_table`.

One possible solution is to combine two of those options. Get the `make_diag_table` process spit out a `drag_table` and a mapping doc of variable name to realm. So the two are consistent. Maybe the mapping doc just list variables to override. Ideally it would be something `addmeta` could use directly.

blimlim: > One possible solution is to combine two of those options. Get the make_diag_table process spit out a drag_table and a mapping doc of variable name to realm. So the two are consistent. Maybe the mapping doc just list variables to override.

I think my preference would be to create a mapping doc by hand and keep it static, as I think the full list of possible wombatlite variables that could be saved in ESM1.6 would remain almost static, and so dynamically writing the mapping doc might just result in subsets of this list being produced.

Currently `make_diag_table.py` is quite flexible about the structure of the `diag_table_source.yaml` files and by requiring it to specially treat a new `realm` setting in the files might lose some of that flexibility. 

Let me know if you have any thoughts on this!

blimlim:
Apart from swapping to a release `model-processing` environment, this is ready to go.

One year of output generated with the latest changes is available in `/scratch/tm70/sw6175/access-esm/archive/all-mom-ocean-313-data-spec-3265c7b8`

scripts/post-processing/addmeta/dataspec.yaml
```diff
@@ -0,0 +1,61 @@
+global:
+    base_configuration: "dev-preindustrial+concentrations"
```
aidanheerdegen: Should this be a `release-` config?

scripts/post-processing/addmeta/dataspec.yaml
```diff
@@ -0,0 +1,61 @@
...
+    contact: "{{ metadata.contact }} ({{ metadata.email}})"
+    Conventions: "CF-1.11,ACDD-1.3"
+    data_specification: "ACCESS-Output Data Specification v0.1.0-alpha"
```
aidanheerdegen: Latest release is [0.3.0](https://github.com/ACCESS-NRI/access-output-data-specifications/releases/tag/v0.3.0)

scripts/post-processing/addmeta/dataspec.yaml
```diff
@@ -0,0 +1,61 @@
...
+    date_created: "{{ __file__.mtime }}"
+    date_metadata_modified: "{{ __datetime__.now }}"
+    # date_modified refers to data but we just modified the files
+    date_modified: "{{ __datetime__.now }}"
+    experiment_repo: "{{ metadata.url }}"
+    experiment_uuid: "{{ metadata.experiment_uuid }}"
+    # FIXME: Is filename one needed? It's not needed by the spec, check with Aidan
+    filename: "{{ __file__.name }}"
```
aidanheerdegen: I recall a conversation at one stage that it was useful to include this at the point of creation (or close to it) of the file, as files can be renamed and this at least records the original filename.

I suppose if the metadata is embedded in the file netCDF metadata maybe this is no longer relevant?

scripts/post-processing/addmeta/ocean.yaml
aidanheerdegen: So it isn't required?

scripts/post-processing/post_process.sh
```diff
@@ -0,0 +1,48 @@
...
+do
+    addmeta \
+        -v -s \
+        -d metadata.yaml \
+        -d $PAYU_CURRENT_OUTPUT_DIR/env.yaml \
+        -m scripts/post-processing/addmeta/dataspec.yaml \
+        -m scripts/post-processing/addmeta/${submodel}.yaml \
+        --fnregex='access-esm1p6\.\w+(?:\.\dd)?\.(?P<var>\w+)\.(?P<freq>\w{2,4})(?:\.\w+)?(?:\.\d{4})?\.nc' \
+        $PAYU_CURRENT_OUTPUT_DIR/${submodel}/*.nc
+done
```
aidanheerdegen, line 48: This is nice and clean!

scripts/post-processing/addmeta/dataspec.yaml
```diff
@@ -0,0 +1,61 @@
...
+    contact: "{{ metadata.contact }} ({{ metadata.email}})"
+    Conventions: "CF-1.11,ACDD-1.3"
+    data_specification: "ACCESS-Output Data Specification v0.3.0"
+    date_created: "{{ __file__.mtime }}"
+    date_metadata_modified: "{{ __datetime__.now }}"
+    # date_modified refers to data but we just modified the files
+    date_modified: "{{ __datetime__.now }}"
+    experiment_repo: "{{ metadata.url }}"
+    experiment_uuid: "{{ metadata.experiment_uuid }}"
+    # FIXME: Is filename one needed? It's not needed by the spec, check with Aidan
```
joshuatorrance: Filename isn't part of the dataspec but doesn't do any harm.

```suggestion
```

scripts/post-processing/addmeta/dataspec.yaml
```diff
@@ -0,0 +1,61 @@
...
+    geospatial_lon_max: -180
+    geospatial_lon_min: 180
+    grid: "access-1deg-global"
+    id: "{{ metadata.experiment_uuid }}"
+    license: "{{ metadata.license }}"
+    model: "{{ metadata.model }}"
+    model_version: "{{ metadata.version }}"
+    parent_experiment_branch_time: "{{ metadata.parent_experiment_branch_time }}"
+    parent_experiment_uuid: "{{ metadata.parent_experiment }}"
+    run_id: "{{ env.PAYU_RUN_ID }}"
```
joshuatorrance, line 23: Is PAYU_RUN_ID the appropriate commit hash to the config repo?

scripts/post-processing/addmeta/dataspec.yaml
```diff
@@ -0,0 +1,61 @@
+global:
+    base_configuration: "dev-preindustrial+concentrations"
+    contact: "{{ metadata.contact }} ({{ metadata.email}})"
+    Conventions: "CF-1.11,ACDD-1.3"
+    data_specification: "ACCESS-Output Data Specification v0.3.0"
+    date_created: "{{ __file__.mtime }}"
+    date_metadata_modified: "{{ __datetime__.now }}"
+    # date_modified refers to data but we just modified the files
```
joshuatorrance:
```suggestion
```

scripts/post-processing/addmeta/dataspec.yaml
```diff
@@ -0,0 +1,61 @@
...
+    date_modified: "{{ __datetime__.now }}"
+    experiment_repo: "{{ metadata.url }}"
```
joshuatorrance, line 9: Does this resolve to the appropriate config repo? This + run_id should identify an appropriate commit to identify the config used for the experiment.

scripts/post-processing/post_process.sh
```diff
@@ -0,0 +1,48 @@
...
+then
+    splitnc --shared-vars uarea,tmask,tarea --excluded-vars VGRD.  --use-esm1p6-filenames --fix-cell-methods --skip-existing "${icefiles[@]}"
+    if [ $? -eq 0 ]; then
+      rm "${icefiles[@]}"
+    fi
+fi
+
+
+# Clean up global metatdata to meet ACCESS-NRI dataspec standards
+# https://access-output-data-specifications--2.org.readthedocs.build/en/2/
```
joshuatorrance:
```suggestion
# https://access-output-data-specifications.readthedocs.io/en/latest/specification/
```

scripts/post-processing/addmeta/dataspec.yaml
aidanheerdegen: *Ideally* yes. It does depend on it being set correctly manually in the experiment however.

scripts/post-processing/post_process.sh
blimlim: We are now using `ocean` as the realm for all MOM output variables, renaming is no longer needed.

scripts/post-processing/post_process.sh
blimlim: Tidied up now, especially with the separate realm step removed!

scripts/post-processing/addmeta/dataspec.yaml
blimlim: Thanks, updated!

scripts/post-processing/addmeta/ocean.yaml
blimlim: This file has now been updated so that it uniformly applies the `ocean` realm to all MOM files

scripts/post-processing/addmeta/dataspec.yaml
blimlim: It will point to the appropriate hash if the user uploads their experiment to a repository and sets the `metadata.url` field to that repository. Otherwise it won't automatically be accessible.

@aidanheerdegen, is that correct?

scripts/post-processing/addmeta/dataspec.yaml
blimlim: Yes good point! I've updated this and added an [issue](https://github.com/ACCESS-NRI/model-config-tests/issues/246) to add a QA test for this as it's something I'm likely to forget.

scripts/post-processing/post_process.sh
blimlim: This has now been updated to the released environment – just want to check whether it should be pinned to a specific version here?

scripts/post-processing/addmeta/dataspec.yaml
joshuatorrance: Do I need to change `base_configuration` for the postprocessed datasets I'm preparing in `jq44`?

dougiesquire (commented):
Thanks @blimlim. The changes in the `ocean/` directory all look correct to me

scripts/post-processing/post_process.sh
blimlim: I've pinned it to `model-processing/1.0.0_0` just in case any breaking changes occur that we don't want to automatically get picked up by the configurations

joshuatorrance (approved):

scripts/post-processing/addmeta/dataspec.yaml
blimlim: Thanks, I've just cut out the `filename` setting - happy to add it back in if we change our mind

blimlim:
@aidanheerdegen, this one should be ready to go. If you're happy to take another look at it that would be great!

scripts/post-processing/addmeta/dataspec.yaml
```diff
@@ -0,0 +1,58 @@
+global:
+    base_configuration: "release-piControl"
+    contact: "{{ metadata.contact }} ({{ metadata.email}})"
+    Conventions: "CF-1.11,ACDD-1.3"
+    data_specification: "ACCESS-Output Data Specification v0.3.0"
```
joshuatorrance: I'd forgotten to update the schema version when I last updated the specification.
```suggestion
    data_specification: "ACCESS-Output Data Specification v0.4.0"
```

blimlim:
Thanks @joshuatorrance, @aidanheerdegen and @dougiesquire for reviewing this. I'll cherry pick this into the other configurations, and update the `base_configuration` for each.

blimlim:
!cherry-pick  0acdf8c  into dev-esm-piControl dev-historical dev-esm-historical dev-scen7-h dev-scen7-vl dev-amip

